PR #66, opened 2013-07-19
[COOK-3358] Fix race condition caused by reload of postgresql.conf when bad data is introduced into pg_hba.conf

rhass:
https://tickets.opscode.com/browse/COOK-3358

This patch changes the notifies to use the default delayed action to ensure both config files are on disk before attempting to reload.

### DIFF
--- a/recipes/server.rb
+++ b/recipes/server.rb
@@ -62,7 +62,7 @@ template "#{node['postgresql']['dir']}/postgresql.conf" do
   owner "postgres"
   group "postgres"
   mode 0600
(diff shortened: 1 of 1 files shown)


slyness:
This has been resolved as of 3.4.12 
Functionality is refactored into recipe server_conf.rb:
https://github.com/hw-cookbooks/postgresql/blob/develop/CHANGELOG.md#v3412 
 
Commit: 
https://github.com/hw-cookbooks/postgresql/commit/57ae1cb8cc2f8076eb93026a329f6a24c1aa7f20

PR #82, opened 2013-10-04
[COOK-3753] Fix dollar sign in password bug

victorlin:
When there is a dollar sign in password, it will be set wrongly (interpreted as environment variable). For example

```
echo "ALTER ROLE postgres ENCRYPTED PASSWORD 'MY_SUPER_$PWD';"
```

will become

```
ALTER ROLE postgres ENCRYPTED PASSWORD 'MY_SUPER_/path/to/pwd';
```

### DIFF
--- a/recipes/server.rb
+++ b/recipes/server.rb
@@ -84,7 +84,7 @@ end
 bash "assign-postgres-password" do
   user 'postgres'
   code <<-EOH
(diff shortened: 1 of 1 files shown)


slyness:
No activity since 2013. PR is no longer applicable to Master.

PR #107, opened 2014-01-23
add support for installing plpython and plpython3 extensions

kisoku:
very trivial support for installing plpython and plpython3 extensions, 

is this even worth adding additional suites for in the TK config ?


slyness:
@kisoku @michaelkirk We are going to close this PR. If you feel like there is a valid reason to still discuss adding plpython in this manner please let us know.

Issue #147, opened 2014-05-23
postgresql::server fails on Debian Wheezy


infertux:
Output from Chef:

```
================================================================================
Error executing action `create` on resource 'template[/etc/postgresql/9.1/main/postgresql.conf]'
================================================================================


Chef::Exceptions::EnclosingDirectoryDoesNotExist
------------------------------------------------
Parent directory /etc/postgresql/9.1/main does not exist.


Resource Declaration:
---------------------
# In /root/chef/cookbooks/postgresql/recipes/server.rb

 62: template "#{node['postgresql']['dir']}/postgresql.conf" do
 63:   source "postgresql.conf.erb"
 64:   owner "postgres"
 65:   group "postgres"
 66:   mode 0600
 67:   notifies change_notify, 'service[postgresql]', :immediately
 68: end
 69: 



Compiled Resource:
------------------
# Declared in /root/chef/cookbooks/postgresql/recipes/server.rb:62:in `from_file'

template("/etc/postgresql/9.1/main/postgresql.conf") do
  provider Chef::Provider::Template
  action "create"
  retries 0
  retry_delay 2
  guard_interpreter :default
  path "/etc/postgresql/9.1/main/postgresql.conf"
  backup 5
  atomic_update true
  source "postgresql.conf.erb"
  cookbook_name :postgresql
  recipe_name "server"
  owner "postgres"
  group "postgres"
  mode 384
end
```

Debug info from the server:

```
root@server:~# cat /etc/debian_version 
7.5
root@server:~# dpkg -s postgresql-9.1
Package: postgresql-9.1
Status: install ok installed
Priority: optional
Section: database
Installed-Size: 16450
Maintainer: Debian PostgreSQL Maintainers <pkg-postgresql-public@lists.alioth.debian.org>
Architecture: amd64
Version: 9.1.13-0wheezy1
Replaces: postgresql-contrib-9.1 (<< 9.1~beta1-3~), postgresql-plpython-9.1 (<< 9.1.13-0wheezy1)
[...]

PR #175, opened 2014-09-19
Create Directory for default 9.1 install under debian (issue #147)

sinfomicien:
- And add unit tests to verify that it works.

### DIFF
--- a/recipes/server_debian.rb
+++ b/recipes/server_debian.rb
@@ -30,3 +30,10 @@ service "postgresql" do
   supports :restart => true, :status => true, :reload => true
   action [:enable, :start]
 end
+
+unless ::File.directory?('/etc/postgresql/' + node['postgresql']['version'] + '/main')
+  execute 'Set locale and Create cluster' do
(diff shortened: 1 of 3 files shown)


recipes/server_debian.rb
```diff
@@ -30,3 +30,10 @@
   supports :restart => true, :status => true, :reload => true
   action [:enable, :start]
 end
+
+unless ::File.directory?('/etc/postgresql/' + node['postgresql']['version'] + '/main')
```
martinisoft: So you know, this can be removed in favor of using a not_if for the execute resource, see https://docs.chef.io/resource_common.html#not-if-examples for some examples.

slyness: Thanks. @sinfomicien can you review @martinisoft's suggestion and update or respond accordingly. I would love to merge this soon.

slyness:
I have extended your PR and included the change to use the not_if guard with PR #222 . Thank you!

PR #200, opened 2014-10-14
If not assigning a password, skip chef-solo check and node.save

aclowes:
This change wraps the chef-solo postgres password check and the code that sets the password attribute in an if statement: if we are not setting the postgres password, skip these operations.

Ideally I think we should get rid of both these operations:
- The chef-solo check is not needed because the assign-postgres-password  resource always runs, does not check if the password has changed, and does not need authentication itself. Therefore it shouldn't matter if we are changing the password with every chef-solo run.
- The node.save() statement is unusual behavior ever since chef 10, when they switched to only saving node attributes if the run was successful. I don't think this use case warrants disregarding the convention, and if so it should be feature flagged.

I've done the minimal change but if others suggest a more extensive change, such as not setting the password by default or removing the save and solo check completely, I'd be happy to do those.

### DIFF
--- a/recipes/server.rb
+++ b/recipes/server.rb
@@ -20,29 +20,33 @@
 include_recipe "postgresql::client"
 
 # randomly generate postgres password, unless using solo - see README
-if Chef::Config[:solo]
-  missing_attrs = %w{
-    postgres
-  }.select do |attr|
-    node['postgresql']['password'][attr].nil?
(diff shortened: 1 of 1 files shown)


slyness:
We will review this. Thanks for the updates.

damacus:
Thanks, I think we'll find a similar way to do this in the updated cookbook!

PR #201, opened 2014-10-17
Define version-sensitive attributes in a recipe

joshk0:
This allows the postgresql cookbook to continue working after default
    attributes are used to set the postgresql version. The postgresql
    version is used to create many other attributes containing that version
    number.
    
    By defining these version-sensitive attributes in a recipe, they can
    safely be defined knowing that all attributes files have already been
    loaded and that node[:postgresql][:version] has its final value.
    
    Otherwise, setting a default postgresql version in an attribute file
    outside the cookbook has no coherent effect.
    
    A caveat of this is that you can't override any config_version
    attributes anymore using default attributes defined in attribute files.
    You have to use force_default. But it turns out that most of these
    attributes should not really be changed, except for the postgresql.conf
    fields.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -31,18 +31,6 @@ when "debian"
     default['postgresql']['version'] = "9.1"
   end
 
-  default['postgresql']['dir'] = "/etc/postgresql/#{node['postgresql']['version']}/main"
-  case
-  when node['platform_version'].to_f < 6.0 # All 5.X
-    default['postgresql']['server']['service_name'] = "postgresql-#{node['postgresql']['version']}"
-  else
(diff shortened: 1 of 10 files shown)


bai:
I'd love to see this. :+1:

kri5:
:+1:

slyness:
This is now under review.

mrjcleaver:
As this disappeared in 3.4.18, is the progress on making it re-appear (in a 3.5 release) tracked somewhere?

mvdkleijn:
@slyness Could you please take a look at this? There seems to be a regression in the cookbook causing this merge to have been dropped. This is causing some issues in our cookbook as a result.

PR #219, opened 2015-01-22
Changed hard coded value to attribute

jensjorritsma:
Was utilizing contrib to install an extension and noticed that the recipe defines 'template1' for the postgres database.  To make the adjustable for various use cases, I created a new attribute to replace hard coded instances of 'template1'.


slyness:
Seems quite reasonable. Thanks.

PR #227, opened 2015-02-19
Openssl Dependency Change

slyness:
Reverts hw-cookbooks/postgresql#226


slyness:
We would like to pin to previous version that most systems will be using. And push the major change to the dependency cookbook into a new version release later.

slyness:
This will be covered with the 3.4.18 release. #231

PR #228, opened 2015-02-20
pin openssl cookbook to version due to upstream major version change

slyness:


cwjohnston:
I think this looks good.

PR #230, opened 2015-02-23
Support for server installation with systemd in RHEL based distributions

pbanderas:
Use of postgresql-setup should be applied to all linux distributions using systemd (RHEL 7, for example). Otherwise installing PostgreSQL in Centos 7.0, among others, will fail (as it is not allowed to user service with initdb); in fact, at least in my case, kitchen test in this plaftform was failing.

I used the node attribute "init_package", I think it's more appropiate. 

Also I noticed that postgresql-setup is not always available in PATH, at least in PostgreSQL 9.3, even the script name is different. So I added a new attribute for this 'setup_script'. Maybe an alternative is the more general **initdb** script located in the same bin dir. I believe it's common on all platforms. I have not tried this approach however.

### DIFF
--- a/recipes/config_version.rb
+++ b/recipes/config_version.rb
@@ -61,6 +61,12 @@ when "fedora"
   node.default['postgresql']['contrib']['packages'] = %w{postgresql-contrib}
   node.default['postgresql']['server']['service_name'] = "postgresql"
 
+  if node['postgresql']['version'] == '9.3'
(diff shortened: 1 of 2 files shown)


slyness:
Can you please rebase this PR to latest develop?

pbanderas:
Due to my remarkable lack of skills on github, I've been unable to use this same PR
I've opened a new one at https://github.com/hw-cookbooks/postgresql/pull/247

PR #231, opened 2015-02-23
Release 3.4.18

slyness:
Release covers two main issues. 

This reverts changes from #201 which will move some attribute resources. These changes will likely be re-included in a more significant version release. 

This also addresses upstream openssl cookbook release issues.


webframp:
Looks bueno. :+1:

slyness:
Thanks.

legal90:
@slyness Could you tell please, what is the reason of reverting #201 ? 
Now the version overriding is painful again :(

slyness:
After releasing 3.4.16 we heard from other users that the changes #201 were causing sudden and unexpected problems with their own use of this cookbook. As a result, we decided that the best course of action was to revert the change in the current 3.x branch and revisit it as part of a future 4.0 release.
Because we're trying to follow [SemVer](http://semver.org), we should probably have realized #201 was a potentially breaking change and held off on merging it.

We hope to get these change back in again soon.

legal90:
@slyness OK, got it. But could you please refer me to the issues which #201 caused? I just thinking about applying the same changes to some other cookbooks, so I'd like to see what can be broken.

"Derived attributes" is big problem of Chef (explained here: https://coderanger.net/derived-attributes/). Current state of `postgresql` cookbook doesn't work as expected. 
Even here is an unreachable code:
https://github.com/hw-cookbooks/postgresql/blob/develop/attributes/default.rb#L119-L120

cwjohnston:
@legal90 at least one of the issues reported to us with regard to #201 was received via an IRC conversation with @kisoku, so I don't think we have a github issue or other ticket to refer you to. 

It's possible that the changes introduced in #201 only impact folks who were doing unforeseen things to modify the existing node attributes for reasons unrelated to overriding the version, but regressions of this sort are the kind of thing semantic versioning is supposed to help us avoid. 

I agree that derived attributes are a problem, and addressing that in this cookbook is a good idea, but needs to be done as part of a major version change.

legal90:
@cwjohnston OK, thanks. I'm looking forward to the new major release.

patcon:
Can we revisit this? I worry that it's a bad sign that we haven't worked to get rid of these derived attributes...

PR #242, opened 2015-03-30
Only save postgresql password to node attributes when it's actually set

chr4:
This prevents random passwords taining node configuration attributes


damacus:
Thanks for this. We'll look at doing it in a different way. 

We dont' want to be doing node.save anymore as it's not recommended behaviour. We're also moving the cookbook to custom resources.

PR #244, opened 2015-03-31
adds ident map override feature

axos88:
This will enable to specify idents as an overridable map


tas50:
I'm going to close this PR out at this point. We've removed all the existing recipe code and we're working to rewrite things using resources and the official postgresql repositories. Checkout what we have now and certainly feel free to shoot us some feedback in Slack

PR #247, opened 2015-04-07
Support for server installation with systemd in RHEL based distributions

pbanderas:
Considers the ohai attribute 'init_package' to establish service name and setup script (cannot run `service initdb` as it's not allowed with systemd)
This setup script changes location (and it's not included in $PATH) whether it comes from the native distro yum repo (default version), or pgdg repo.

This are the changes over the develop branch. I've reduced them to consider only version 9.3 from yum pgdg repos enabled, as the kitchen tests specify. Verified with all server-centos and yum-pgdg-centos kitchen machines.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -22,6 +22,51 @@ default['postgresql']['assign_postgres_password'] = true
 # Establish default database name
 default['postgresql']['database_name'] = 'template1'
 
+# Sets OS init system (upstart, systemd, ...), instead of relying on Ohai
+default['postgresql']['server']['init_package'] = case node['platform']
+  when 'debian'
+    case
+    when node['platform_version'].to_f < 7.0
+      'sysv'
+    else
+      'systemd'
+    end
+  when 'ubuntu'
+    case
+    when node['platform_version'].to_f < 15.04
+      'upstart'
+    else
+      'systemd'
+    end
+  when 'amazon'
+    'upstart'
+  when 'redhat', 'centos', 'scientific', 'oracle'
+    case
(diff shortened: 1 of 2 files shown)


stiff:
This looks like a very helpful PR. Waiting for it a lot.

chrissexton:
I've been using this branch for a bit, it'd be really nice to get it merged.

patcon:
Trying to use this, but might not work on docker. As for as I can tell`node['init_package']` is `sshd` in dockerized images

patcon:
Yeah, seems that detecting init systems is very finicky:
http://unix.stackexchange.com/questions/18209/detect-init-system-using-the-shell

We should probably just assume RHEL 7 and CentOS 7 are using the default systemd for now, and if someone needs to generalize the logic later, we can let them do that. Would that work?

pbanderas:
You're right the Ohai init_package is not as accurate as it should be, there was a discussion about this PR: https://github.com/chef/ohai/pull/301, last comment shows an example using a mapping.

I could rewrite the PR so it gets the init system from such a mapping, like in here https://github.com/jordansissel/pleaserun/blob/master/lib/pleaserun/detector.rb#L10-L32
Result should be exactly the same.

slyness:
If you could please rebase this sometime I will try and get it tested and merged.

pbanderas:
Done. I've added a new attribute to establish init system by hand, instead of using Ohai.
Also included the svc_name variable after initdb as in https://github.com/hw-cookbooks/postgresql/blob/develop/recipes/server_redhat.rb#L98

PR #253, opened 2015-05-04
Removed hardcoded UID/GID for RHEL systems

alexpekurovsky:
For RHEL systems, where uid/gid 26 already used by another applications, uid/gid configuration available from attributes

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -73,6 +73,9 @@ when "ubuntu"
 
 when "fedora"
 
+  default['postgresql']['uid'] = "26"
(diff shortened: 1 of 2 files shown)


kesor:
:+1:

nbugrovsky:
thanks, i was waiting for it :+1:

slyness:
You're replacing an integer with a string here.. Do you have a reason for doing that?

kesor:
+1 about the int -> string change. although it doesn't actually matter in chef.
https://github.com/chef/chef/blob/master/lib/chef/resource/user.rb#L75

PR #256, opened 2015-05-09
Amazon linux 2015 pgdg support

patarra:
The recipe yum_pgdg_postgresql throws an error "key 2015 not exists" when trying to converge in a ec2 amazon linux 2015 instance, as there is no definition for the rpm url of pgdg yum repositoriy.


slyness:
covered with #240 merge. Also, there is no need to include version changes to the cookbook version. PRs to this repository are made to the develop branch and often we will include multiple updates per release. Thanks for the input and contributions.

patarra:
Thanks Aaron, good to know!!

PR #259, opened 2015-05-18
node.save could better not be run on every chef 

kstruis:
node.save are costly and could better not be run on every chef run since it causes node.default attributes stored to the node objects to differ during a chef run and when the node finally saves.

In other words, in-between node.save calls will cause attribute defaults set from subsequent cookbook recipes to be removed from the object, until the chef-run is complete and saved again.

In a scenario we are running this causes problems. Hope this can be merged.


slyness:
Indeed. Thanks.

PR #262, opened 2015-05-22
restart postgres service immediately on config change

reset:
Dupe of https://github.com/hw-cookbooks/postgresql/pull/255 but against develop branch instead of master


slyness:
Awesome. Thanks.

PR #264, opened 2015-05-25
Don't use PGDG for 9.4 on Debian

chewi:
I realise there was a small mistake in my earlier pull request (it broke Wheezy) so you used someone else's but they missed the line about which versions to enable PGDG for.

### DIFF
--- a/recipes/client.rb
+++ b/recipes/client.rb
@@ -19,7 +19,7 @@ include_recipe 'postgresql::ca_certificates'
 
 case node['platform_family']
 when 'debian'
-  if node['postgresql']['version'].to_f > 9.3
+  if node['postgresql']['version'].to_f > 9.4
     node.normal['postgresql']['enable_pgdg_apt'] = true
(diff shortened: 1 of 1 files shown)


rmoriz:
why not using PGDG?

chewi:
Because 9.4 is in Jessie as well as Stretch. It may end up being in Wheezy Backports as well. To be honest, the logic here isn't great because it doesn't take account of which Debian version you're actually running. Even still, I don't think it should try to enable PGDG by default unless you've requested a version that we know isn't available in any Debian release yet.

chewi:
I have rebased this. Please? It's just one line!

chewi:
Once again…

slyness:
It's not you. It's time. I'll try and get to it soon.

chewi:
That's okay, it needed rebasing anyway.

chewi:
:tada: :tada: :tada:

PR #268, opened 2015-05-28
Fix specs

cwjohnston:
For some reason we had two different spec dirs and no defined way to run one of them. This change unifies those specs under `tests/unit`, fixes the specs broken due to unstubbed commands, and adds rake tasks for running rspec tests, test-kitchen! :eggplant:


luckymike:
:+1: Not comprehensiveness of this test coverage, but :vertical_traffic_light: is all green!

PR #269, opened 2015-05-29
Better CentOS 5, 6, 7 support

merqlove:
- Added 9.4 rpms for CentOS 5,6.
- Fixed CentOS Init for CentOS 7+.
- Fixed Architecture check (For wrong response).

### DIFF
--- a/recipes/server_redhat.rb
+++ b/recipes/server_redhat.rb
@@ -83,7 +83,7 @@ elsif platform?("redhat") and node['platform_version'].to_i >= 7
     not_if { ::FileTest.exist?(File.join(dir, "PG_VERSION")) }
   end
 
-elsif platform?("centos") and node['platform_version'].to_i >= 6
+elsif platform?("centos") and node['platform_version'].to_i >= 7
 
   execute "/usr/pgsql-#{node['postgresql']['version']}/bin/postgresql#{node['postgresql']['version'].split('.').join}-setup initdb #{svc_name}" do
     not_if { ::FileTest.exist?(File.join(dir, "PG_VERSION")) }


ozzyogkush:
I've run this PR on my project with postgres 9.3 on both centos6.5 and centos6.6 and neither will successfully converge. It still

ozzyogkush:
I've run this PR on my project with postgres 9.3 on both centos6.5 and centos6.6 and neither will successfully converge. It still gives the `Error executing action `restart` on resource 'service[postgresql]'` error

merqlove:
Can you post your vars & recipe list?
Also you have loading `yum_pgdg_postgresql` recipe or its attribute `enable_pgdg_yum` is true?

merqlove:
Ah, i see error. Seems that something in config.
You have looked into postgresql log?

ozzyogkush:
Attributes:

```
include_attribute 'postgresql'

# Set PostgreSQL data
node.default['postgresql']['version'] = '9.3'
case node['platform_family']
  when 'ubuntu'
    node.default['postgresql']['enable_pgdg_apt'] = true
  when 'redhat', 'rhel', 'centos'
    node.default['postgresql']['enable_pgdg_yum'] = true
end
node.default['postgresql']['config']['listen_addresses'] = '*'
node.default['postgresql']['contrib']['packages'] = [
    'pgadmin3'
]
node.default['postgresql']['contrib']['extensions'] = [
    'adminpack'
]
```

ozzyogkush:
The recipe runs `include_recipe 'postgresql::server'`, then simply creates a DB user and a DB. The DB resource call includes `notifies :restart, "service[#{node['postgresql']['server']['service_name']}]", :immediately`, which is where this fails.

merqlove:
I have this one in my cookbook db::provider on install step:

case node['platform_family']
  when 'debian', 'ubuntu'
    node.default['postgresql']['enable_pgdg_apt'] = true
  when 'rhel', 'fedora', 'suse'
    node.default['postgresql']['enable_pgdg_yum'] = true
  end

  run_context.include_recipe 'postgresql::contrib'

  run_context.include_recipe 'postgresql::config_initdb'
  run_context.include_recipe 'postgresql::config_pgtune'

  run_context.include_recipe 'postgresql::ruby'

ozzyogkush:
I'll try your suggestions shortly.

merqlove:
Also you can try to make provider specially for DB. In provider you can run such things like:
`run_context.include_recipe 'postgresql::contrib'```
Not in runtime.

ozzyogkush:
Tried yours, same result :(

merqlove:
@ozzyogkush Can you look into Postgresql log?(data folder)

ozzyogkush:
Just calling `postgresql::server` gets me:

```
[vagrant@database-centos-65 ~]$ sudo cat /var/lib/pgsql/data/pg_log/postgresql-Sat.log
LOG:  database system was shut down at 2015-05-30 22:59:04 UTC
LOG:  database system is ready to accept connections
LOG:  autovacuum launcher started
LOG:  received fast shutdown request
LOG:  aborting any active transactions
LOG:  autovacuum launcher shutting down
LOG:  shutting down
LOG:  database system is shut down
LOG:  invalid authentication method "peer"
CONTEXT:  line 28 of configuration file "/var/lib/pgsql/data/pg_hba.conf"
FATAL:  could not load pg_hba.conf
LOG:  invalid authentication method "peer"
CONTEXT:  line 28 of configuration file "/var/lib/pgsql/data/pg_hba.conf"
FATAL:  could not load pg_hba.conf
```

line 28 of /var/lib/pgsql/data/pg_hba.conf: `local   all             all                                     peer`

merqlove:
Try this config for peer:

```
{
  "postgresql": {
    "pg_hba": [
      {
        "type": "local",
        "db": "all",
        "user": "postgres",
        "addr": null,
        "method": "ident"
      },
      {
        "type": "local",
        "db": "all",
        "user": "all",
        "addr": null,
        "method": "ident"
      },
      {
        "type": "host",
        "db": "all",
        "user": "all",
        "addr": "127.0.0.1/32",
        "method": "md5"
      },
      {
        "type": "host",
        "db": "all",
        "user": "all",
        "addr": "::1/128",
        "method": "md5"
      }
    ]
  }
}
```

ozzyogkush:
https://github.com/hw-cookbooks/postgresql/blob/develop/templates/default/pg_hba.conf.erb automatically includes that line:

```
<% elsif node['postgresql']['version'].to_f >= 9.1 -%>
local   all             all                                     peer
<% end -%>
```

merqlove:
Just do it with overriding attribute :) Higher priority.

ozzyogkush:
I'm not sure what you mean by that. I'm installing PG 9.3, so that line will always run.

I add to the pg_hba attribute like so:
```
node.default['postgresql']['pg_hba'].push(
    {
        :type => 'host',
        :db => 'all',
        :user => 'all',
        :addr => '10.0.2.2/32',
        :method => 'md5'
    },
    {
        :type => 'host',
        :db => 'all',
        :user => 'all',
        :addr => '10.11.22.0/24',
        :method => 'md5'
    }
);
```

merqlove:
I'm use for this: node.override. Because this is array.

merqlove:
So as i see now, you have trouble with configuration of pg_hba.

ozzyogkush:
My method works just fine. Final output of `/var/lib/pgsql/data/pg_hba.conf` is 

```
[vagrant@database-centos-65 ~]$ sudo cat /var/lib/pgsql/data/pg_hba.conf
# This file was automatically generated and dropped off by Chef!

# PostgreSQL Client Authentication Configuration File
# ===================================================
#
# Refer to the "Client Authentication" section in the PostgreSQL
# documentation for a complete description of this file.

# TYPE  DATABASE        USER            ADDRESS                 METHOD

###########
# Other authentication configurations taken from chef node defaults:
###########

local   all             postgres                                ident

local   all             all                                     ident

host    all             all             127.0.0.1/32            md5

host    all             all             ::1/128                 md5

host    all             all             10.0.2.2/32             md5

host    all             all             10.11.22.0/24           md5

# "local" is for Unix domain socket connections only
local   all             all                                     peer
```

merqlove:
Haha, glad to hear! :+1: Cheers!

ozzyogkush:
Yes it's definitely a problem with that, hence `LOG:  invalid authentication method "peer"` <--- this is what I don't get though. I'm looking through the PG docs and it definitely is a valid method for 9.3 :/

ozzyogkush:
Ok, I've checked and its actually only installing 8.4 of postgresql, which makes sense. I figured out why it's doing it - derived attributes! Even though I set the version to 9.3 in my attributes file, it's already creating the `default['postgresql']['server']['packages'] = ["postgresql#{node['postgresql']['version'].split('.').join}-server"]` attribute based on the original version defined in this cookbook. It needs to use lazy attributes, using the % operator instead of #

merqlove:
Any way only this way works within CentOS 7 & SystemD

zocimek:
Is this  change going to be merged ?

ozzyogkush:
That's up to the owners of the repo https://github.com/orgs/hw-cookbooks/people, but I'd hope so.

PR #275, opened 2015-06-19
add attribute to use templates from other cookbook in server_conf.rb recipe

carltonstedman:
Added attributes which default to `nil`.

When not `nil`, allows using templates in recipes/server_conf.rb (e.g. for postgresql.conf, pg_hba.conf) from another cookbook.

This is simpler than requiring the use of chef-rewind and allows for overriding the templates when postgresql community cookbook is included in a LWRP (e.g. with `run_context.include_recipe`), where `rewind` may not behave as expected.

Intent to address https://github.com/hw-cookbooks/postgresql/issues/274

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -220,6 +220,11 @@ default['postgresql']['enable_pgdg_yum'] = false
 
 default['postgresql']['initdb_locale'] = nil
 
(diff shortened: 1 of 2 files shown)


damacus:
Going to close this one down now. 

We'll keep the functionality in mind when we update the cookbook. 

Thanks!

PR #277, opened 2015-06-29
openssl secure password generation issue

GitBytes:
I noticed in the README that the OpenSSL call to generate a secure password for the postgresql database in this cookbook was broken. I think this may fix it. The server recipe seemed to be using an outdated feature that will go away with a future release of the openssl cookbook. Please feel free to test and verify this method. Thanks.  Best Regards, GitBytes (Ian)

### DIFF
--- a/recipes/server.rb
+++ b/recipes/server.rb
@@ -42,7 +42,7 @@ else
   # useful if it weren't saved as clear text in Chef Server for later
   # retrieval.
   unless node.key?('postgresql') && node['postgresql'].key?('password') && node['postgresql']['password'].key?('postgres')
-    node.set_unless['postgresql']['password']['postgres'] = random_password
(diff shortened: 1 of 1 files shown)


bottkv488:
+1 for helping getting rid of the
depends 'openssl', '~> 4.0.0'
contraint

As the random_password method is slightly different from secure_password we might consider using :base64 instead of :hex (random_password(length: 20, mode: :base64))?

Regards, Bernhard

GitBytes:
@bottkv488 Just letting you know I have made the adjustment you recommended. Thanks for the suggestion!

Best Regards, GitBytes (Ian)

bottkv488:
@GitBytes I'm glad you liked it :)

GitBytes:
@slyness Awesome! Thanks for doing that!

PR #279, opened 2015-07-07
Postgresql YUM pgdg only has fedora and redhat platform.  Enable yum …

barrowkwan:
When using this cookbook with yum pgdg enable on Centos, this will break as Postgresql yum pgdg only has redhat and fedora.  Running this on Centos will get KeyError.  Force to fetch "redhat" instead of "centos" will fix this issue

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -357,7 +357,7 @@ end
 def pgdgrepo_rpm_info
   repo_rpm_url = node['postgresql']['pgdg']['repo_rpm_url'].
     fetch(node['postgresql']['version']).            # e.g., fetch for "9.1"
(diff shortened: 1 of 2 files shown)


barrowkwan:
this seem to break as well if not using yum pgdg.  I will submit another pull request

barrowkwan:
latest commit should fix issue if yum pgdg is not enabled

barrowkwan:
postgresql94-setup initdb is broken with pgdg enable as it installed some binaries under /usr/pgsql-9.4/bin.

barrowkwan:
I am going to close this pull request as I found out adding centos repos location in the default.rb will fix this problem.  I will submit a new pull request..

PR #285, opened 2015-08-21
Fixed issue 248. execute call was passing in more than MAJOR and MINO…

wojnosystems:
…R version numbers to the pg_createcluster script, causing the version to be read as blank. Parsed the version and stripped off any numbers except the first two.


wojnosystems:
This actually breaks the not_if command, which I missed. I've fixed what I broke in the second commit. Tested and working on Debian Wheezy.

wojnosystems:
I need to work on this some more, I'll post back when it's complete.

PR #287, opened 2015-09-07
added ability to set locale on pg_createcluster

pauleames:
The postgresql-9.3 package on Ubuntu creates a 'main' cluster.

This pull request allows us to configure the main cluster ourselves, by dropping and re-creating it.
Idempotence is achieved by checking if the 'main' cluster is already installed prior to running the package install.


pauleames:
I'll assume from the lack of discussion that this is unappealing and I'll just create a new cluster instead.

PR #289, opened 2015-10-12
Fix server name references. WIth PostgreSQL, the init.d service is na…

pmoghaddam:
…med "postgresql-9.4" and thus any early restarts via `notifies :restart, 'service[postgresql]'`. Currently the logic has the Chef service resource definition occurring after the restarts, which leads to failures. This fix is to resolve the order.

Note: please double check in case I am misunderstanding how the logic is suppose to work.


damacus:
This is now fixed as part of the server_install resource. Thanks!

PR #290, opened 2015-10-14
fix minitest-handler from kitchen runlist.…

slyness:
see [TESTING](../TESTING.md) for updated testing information

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -1,6 +1,9 @@
 ---
 driver_plugin: vagrant
 
+provisioner:
+  require_chef_omnibus: 12.3
+
 platforms:
 - name: ubuntu-14.04
   run_list:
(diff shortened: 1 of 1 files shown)


cwjohnston:
Given that the existing minitest specs provide better coverage than what we currently have with ChefSpec, I think we should try to retain these tests until we can enhance the ChefSpec and add ServerSpec for integration tests. TESTING.md still applies.

From my testing locally it seems that our usage of minitest-handler breaks under Chef ~> 12.4; specifying the following in .kitchen.yml gets those tests working again:

```
provisioner:
  require_chef_omnibus: 12.3
```

slyness:
Thanks Camero. I'll give that a try.

slyness:
No longer need to make changes. Corrected local testing environment by removing an old crusty Cheffile.lock and converging again.

PR #293, opened 2015-10-26
Improve how postgres password is set

Ginja:
- Changed resource from bash to execute so that the sensitive property can be used. 
- Modified the not_if guard to actually check if it can use the password to make a simply query. This check assumes that pg_hba.conf forces a password for TCP connections. The password comes from the PGPASSWORD environment variable.


slyness:
Hi. Will you please rebase this. I've merged #277 which also makes some updates to the password routines of this recipe. Thanks.

PR #294, opened 2015-10-28
Ported setup_users and setup_databases recipes

sandrew:
I've ported for my purposes recipes for setting up users and databases from phlipper/chef-postgresql cookbook.

Nor sure if you need them in your cookbook, so if you do, here you are.


damacus:
This looks like a great base for a set of resources in the future, thanks so much!

damacus:
Notes for anyone wanting to post this in the the new model. 

- need to make sure the PG gem is installed first
- use custom resources rather than definitions
- could use a bunch of tests

PR #298, opened 2015-11-05
Add Gem.clear_paths to server minitest so pg gem is loaded correctly

slyness:
Add Gem.clear_paths to server minitest so pg gem is loaded correctly on first converge.

Sometimes kitchen test would fail on first run. The test to connect to database would fail because the pg gem wouldn't be available to the chef run.


cwjohnston:
In our testing this one line fix helped to ensure that tests passed on first convergence. :+1:

cwjohnston:
Actually merging

PR #300, opened 2015-11-07
Fix issue where initdb would fail if using pgdg on RHEL7

irvingpop:
The PGDG packages are bad about putting their binaries into the `PATH`.  This patch creates symlinks for the `postgresql94-setup` and `postgresql94-check-db-dir` so that the initdb step doesn't fail with `command not found`


slyness:
Thanks Irving.

PR #308, opened 2015-12-04
add leading pound symbol on comment lines

adamvduke:
The current implementation writes normal lines if a `comment` attribute is provided.


slyness:
Thanks.

PR #318, opened 2016-01-06
Update yum_pgdg_postgresql description to pg version 9.4

treenerd:
Update yum_pgdg_postgresql description to pg version 9.4 and add setup_script attribute for CentOS 7, 7.1.


slyness:
Thanks.

PR #326, opened 2016-01-17
Improve 4.0.0 CHANGELOG

chr4:
As a first step to save others from the pitfalls that I encountered [upgrading to 4.0.0](https://github.com/hw-cookbooks/postgresql/issues/319) , I tried improving the CHANGELOG, so that people have a better understanding of what the changes are doing. I tried including all the issues that I found, but the list might not be complete yet. 

The **WARNING** might be a little too much for your taste, but since not reading it carefully might result in dropped databases it felt like it might be appropriate. Feel free to adapt if necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,49 @@ This file is used to list changes made in each version of the postgresql cookboo
 
 v4.0.0
 -----
+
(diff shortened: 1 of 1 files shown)


slyness:
This seems pretty reasonable addition. If anything it makes it extremely clear to be cautious at this point. I will include it and probably make some additional updates to it for the next release. Thanks for the update.

PR #327, opened 2016-01-18
Include attribute flag for organizations which are using internal copies of the pgdg packages

tduzan:
Hello,

Hope this PR finds you well.  I am requesting this change be included as it is necessary for clean use of this cookbook when wrapped by organizations which maintain their own internal repositories with separate code-signing requirements.  In this case, I am using pgdg packages, but not from the upstream repository, and do not wish to enable it in order to trigger the correct actions to occur.  I have no doubt I am not the only one that might be in this situation.  Best practices say not to fork upstream cookbooks to wrap them, so I am submitting this change to make this possible cleanly.

Thanks.

### DIFF
--- a/recipes/ruby.rb
+++ b/recipes/ruby.rb
@@ -32,11 +32,13 @@ rescue LoadError
   node.set['build-essential']['compile_time'] = true
   include_recipe "build-essential"
 
-  if node['postgresql']['enable_pgdg_yum']
+  if node['postgresql']['enable_pgdg_yum'] || node['postgresql']['use_pgdg_packages']
     package "ca-certificates" do
       action :nothing
     end.run_action(:upgrade)
-
+  end
+  
+  if node['postgresql']['enable_pgdg_yum']
     include_recipe "postgresql::yum_pgdg_postgresql"
 
     rpm_platform = node['platform']
--- a/recipes/ruby.rb
+++ b/recipes/ruby.rb
@@ -32,13 +32,11 @@ rescue LoadError
   node.set['build-essential']['compile_time'] = true
   include_recipe "build-essential"
 
-  if node['postgresql']['enable_pgdg_yum'] || node['postgresql']['use_pgdg_packages']
+  if node['postgresql']['enable_pgdg_yum']
     package "ca-certificates" do
       action :nothing
     end.run_action(:upgrade)
-  end
-  
-  if node['postgresql']['enable_pgdg_yum']
+
     include_recipe "postgresql::yum_pgdg_postgresql"
 
     rpm_platform = node['platform']
--- a/recipes/server_redhat.rb
+++ b/recipes/server_redhat.rb
@@ -53,7 +53,7 @@ node['postgresql']['server']['packages'].each do |pg_pack|
 end
 
 # If using PGDG, add symlinks so that downstream commands all work
-if node['postgresql']['enable_pgdg_yum'] == true
+if node['postgresql']['enable_pgdg_yum'] == true || node['postgresql']['use_pgdg_packages'] == true
   [
     "postgresql#{shortver}-setup",
     "postgresql#{shortver}-check-db-dir"
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -17,6 +17,7 @@
 
 default['postgresql']['enable_pgdg_apt'] = false
 default['postgresql']['enable_pgdg_yum'] = false
+default['postgresql']['use_pgdg_packages'] = false
 
 default['postgresql']['server']['config_change_notify'] = :restart
 default['postgresql']['assign_postgres_password'] = true


luckymike:
Can you expand a little on the use case(s) here? `node['postgresql']['enable_pgdg_yum']` is used for a conditional in 3 recipes (ruby, client, and server_redhat), but you've only adjusted one of them. 

Do you expect the `node['postgresql']['enable_pgdg_yum'] == true` behavior for the ruby and client recipes, but wish to skip it in the server_redhat recipe?

tduzan:
I apologize, my immediate use-case was in the server recipes.  But it should be replicated elsewhere.  I'll update my PR momentarily to incorporate it into the ruby recipe.  The client recipe appears to only use the conditional to check whether or not it should trigger the install of the PGDG repositories, but in this case the point of the extra attribute is when you are using the PGDG packages but not the repository.

Thanks.

tduzan:
My last commit should be all that is required to support this flag if someone wanted to import the ruby recipe.  Everything else should be handled by explicit declarations for other attributes.

Please let me know if there's any other questions/concerns.

Thanks.

luckymike:
Sorry, not trying to be dense. 

>the point of the extra attribute is when you are using the PGDG packages but not the repository

If you're not using the repository, and both these changes do the equivalent of setting `node['postgresql']['enable_pgdg_yum'] = false`, wouldn't you already have the behavior you're after by just setting `node['postgresql']['enable_pgdg_yum']` to `false`?

tduzan:
I'm not sure if I understand your question entirely.  Let me provide a little context on the use case here.  I work in an environment where we host our own internal package repositories, we sign the packages ourselves with an internal key, these are the only repositories and package signing keys allowed to exist on our systems.  Thus, we sometimes use trusted external packages and resign/rehost them, such as in the case of the PGDG PostgreSQL packages.

In server_redhat.rb there are some specific necessary actions taken if you are using the PGDG packages instead of the RHEL-Base packages, because it puts its binaries and other files in a different location (/usr/pgsql-$version/bin/ rather than /usr/bin/ etc).  These actions currently only happen in the upstream cookbook if you set the attribute ```node['postgresql'['enable_pgdg_yum']``` to true.  

In ruby.rb an installation of the "ca-certificates" package is triggered only if you have ```node['postgresql']['enable_pgdg_yum']``` set to true.  My change makes that possible when it's false as long as the new attribute is set.  

Since I'm in a situation where I'm not actually using the PGDG repository, and I cannot have it installed on the server when the cookbook finishes running, I needed a way to support the flexibility of my use case as unobtrusively as possible and I thought it might be helpful to others in a similar situations (internal repos).  If I were to simply set enable_pgdg_yum to true, it would not just install the PGDG repository, but take a significant number of other actions which are undesirable.

Does that provide the context for where this might be warranted?

recipes/server_redhat.rb
```diff
@@ -53,7 +53,7 @@
 end
 
 # If using PGDG, add symlinks so that downstream commands all work
-if node['postgresql']['enable_pgdg_yum'] == true
+if node['postgresql']['enable_pgdg_yum'] == true || node['postgresql']['use_pgdg_packages'] == true
   [
```
tduzan: We still want symlinks created for necessary commands so they'll exist in the $PATH after the cookbook runs, even if we don't set ```node['postgresql']['enable_pgdg_yum']``` to true.

recipes/ruby.rb
```diff
@@ -32,11 +32,13 @@
   node.set['build-essential']['compile_time'] = true
   include_recipe "build-essential"
 
-  if node['postgresql']['enable_pgdg_yum']
+  if node['postgresql']['enable_pgdg_yum'] || node['postgresql']['use_pgdg_packages']
     package "ca-certificates" do
```
tduzan: We still need the ca-certificates package installed if we're using pgdg originated packages without wanting to set the ```node['postgresql']['enable_pgdg_yum']``` attribute to true

luckymike:
Okay, gotcha. I think the correct solution is probably to decouple the client recipe and the ruby/redhat_server recipe actions, rather than doubling up on the attributes, if possible.

tduzan:
Sure, that seems reasonable.  I can think of a few ways to approach that.  Would your preference be for me to try to make that change and submit an update to the PR or to revert my changes to ruby.rb and merge as-is and leave the ruby recipe piece to a separate PR?  My main concern is with the server_redhat.rb change, as we do not use the ruby.rb recipe internally, but I respect that others may need its functionality.

luckymike:
I suggest using your fork for now, as we won't have a new release out immediately. If you want to work on a PR that decouples those, feel free, but we'll also consider this in the current round of updates/fixes.

tduzan:
Alright, that sounds fine.  I'll revert my commit around the ruby.rb changes and we can merge this PR.  I'll try to make something nicer tomorrow after work and submit another PR.

Thanks.

tduzan:
I squashed my typo fix into the original commit and pushed a revert commit.  If you'd like, I can rebase to drop both so it's a single commit PR, but I thought the history might be useful if someone else decides to take a crack at ruby.rb.

Thanks.

tduzan:
@luckymike So was it possible for us to go ahead and merge this initial change?  This deals with the server piece?  We're operating off my fork currently, but would like to work off of upstream if possible.

kemra102:
This is something we need in our organisation as well for more or less the same reasons, getting this merged would be much appreciated.

slyness:
OK. Looks and sounds reasonable. Thanks for contributing.

PR #332, opened 2016-02-15
fix recipes for Ubuntu 16.04 LTS

micred:
The upcoming Ubuntu 16.04 is being shipped with Postgres 9.5.
The postgresql::client recipe is now forcing enable_pgdg_apt when version is > 9.3, so it always force using PDDG APT, even if I just want to use the packaged version already installed with Ubuntu 16.04.

In this PR I removed this check, since I think it is too strict and dependent on which debian-flavoured Linux distribution one is using.


slyness:
thanks, closing in favor of #350

PR #337, opened 2016-02-27
Fix a selection of rubocop violations

madwort:


libraries/default.rb
```diff
@@ -106,9 +104,9 @@ def locale_date_order
 
       if (posM.nil? || posD.nil? || posY.nil?)
         return 'mdy'
-        elseif (posY < posM && posM < posD)
+      elsif (posY < posM && posM < posD)
```
madwort: looks like there was a bug here?

libraries/default.rb
```diff
@@ -210,26 +206,19 @@ def scan_available_timezones(tzdir)
       else # /etc/localtime is a file, so scan for it under tzdir
         localtime_content = File.read("/etc/localtime")
 
-        Find.find(tzdir) do |path|
+        Find.find(tzdir) do |tzpath|
```
madwort: shadows variable in previous `elsif `branch

tas50:
This is being handled in #365

PR #339, opened 2016-03-07
Add 9.5

wgb:
9.5 is not yet in Debian stable, so hold off on bumping tests from 9.4 to 9.5 until then...


chewi:
This has been superseded by #369, which adds 9.5 and 9.6 but also addresses some issues with this pull request:

* Double quotes have since changed to single quotes.
* Fedora 24 has since been released.
* Support for older systems since has been dropped.
* There is no 9.5 RPM for Amazon 2014.
* Oracle 7 and Scientific 7 were missing.
* Amazon and Oracle package strings were wrong.

chewi:
#369 is merged so this can be closed now.

PR #341, opened 2016-03-15
Add checksum to yum pgdg packages

Annih:
IMHO we should always use checksum attributes with remote_file resource; that's why I added the checksum for all pgdg packages.

I also fixed 2 typo for fedora packages.


damacus:
Can you rebase this please? 

Looks good otherwise.

tas50:
I'm going to close this out at this point since we're reworking things to just use the repos directly. That works around this whole problem and the attribute file maint issues that are never ending.

PR #347, opened 2016-03-23
Postgresql 4.0.4

slyness:
* Add leading pound symbol on pg_hba.conf template comment line
* Update gem install for compile_time to correct deprication warning
* Add support Ubuntu Wily Werewolf pgdg apt repository
* test-kitchen platforms for Centos 7.2 and Ubuntu 15.04
* Fixes PostgreSQL version & package name defaults for EL7 distros
* Add appropriate systemd unit file overrides for EL7 distros


agoddard:
:+1: LGTM

PR #350, opened 2016-04-11
Add Support for Ubuntu 16.04

kohenkatz:
According to [the docs](https://wiki.postgresql.org/wiki/Apt), PGDG now has packages for 16.04 Xenial. This should now be allowed to be used here.


slyness:
Thanks for the update.

PR #353, opened 2016-04-25
Support install system packages on Ubuntu 16.04

aashley:
Just a small change so the cookbook can install the system default postgresql packages that come with Ubuntu 16.04.

Also adds a check so that the pgdg repo is not automatically turned on when requesting the ubuntu supplied packages.


tas50:
Added via #360

PR #359, opened 2016-08-10
Add support for chef_zero

smcavallo:
Add support for chef_zero, in addition to chef_solo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@ postgresql Cookbook CHANGELOG
 =============================
 This file is used to list changes made in each version of the postgresql cookbook.
 
-v4.0.7
------
-* Add support for chef zero
-
 v4.0.6
 -----
 * Add 16.04 Xenial to the allowed list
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@ postgresql Cookbook CHANGELOG
 =============================
 This file is used to list changes made in each version of the postgresql cookbook.
 
+v4.0.7
```
tas50: Can you revert this bit. We'll handle the changelog. This confuses people who see it and think that 4.0.7 is pushed.

smcavallo:
@tas50 - Changelog changes have been removed

tas50:
@smcavallo Can you rebase this against the massive dump of changes that I landed on master?

chewi:
What was the point of this change? It worked fine in Chef Zero before because, unlike Chef Solo, it does have somewhere to store the data, under `/tmp/kitchen/nodes`. This change breaks lots of existing Kitchen configurations.

PR #360, opened 2016-08-30
Add support for Ubuntu 16.04

PowerKiKi:
On [Ubuntu 16.04 default version of PostgreSQL is 9.5](http://packages.ubuntu.com/search?keywords=postgresql&searchon=names&suite=all&section=all), but this cookbook was still trying to install 9.3. So default attributes were completed and tests suites were added and verified.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -2,6 +2,10 @@
 driver_plugin: vagrant
 
 platforms:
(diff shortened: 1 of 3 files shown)


mattlqx:
I like this better than https://github.com/hw-cookbooks/postgresql/pull/353 because test!

PR #375, opened 2016-11-15
Update PGDG versions

tas50:


shortdudey123:
This probably needs to be rewritten somehow, but i am fine w/ the version bumps for now

PR #399, opened 2017-01-03
More minor improvements to specs and package installs

tas50:


spec/unit/server_spec.rb
```diff
@@ -35,8 +35,8 @@
           stub_command(/ls \/.*\/recovery.conf/).and_return(false)
         end
 
-        it 'runs no tests' do
-          expect(chef_run)
```
damacus: yolo

tas50: I'm fairly certain the old syntax did the same thing the new one did, but it's more obvious now that we're looking for compilation bugs.

Issue #370, opened 2016-10-29
PostgreSQL 9.6 support


sauliusgrigaitis:
Workaround for PostgreSQL 9.6 support:

```
  "postgresql": {
    "config": {
      "dynamic_shared_memory_type": "sysv"
    },
    "version": "9.6",
    "dir": "/etc/postgresql/9.6/main",
    "client": {
      "packages": ["postgresql-client-9.6","libpq-dev"]
    }, 
    "server": {
      "packages": ["postgresql-9.6"]
    },
    "contrib": {
      "packages": ["postgresql-contrib-9.6"]
    },
  }
```

brauliobo:
+!

nkadel-skyhook:
Postgresql 9.6 has apparently renamed the setup script from "postgresql-setup" to "postgresql96-setup". Some thought, and work, need to be applied to the attributes/default.rb to set this correctly.

aried3r:
> Postgresql 9.6 has apparently renamed the setup script from "postgresql-setup" to "postgresql96-setup".

I cannot verify this, at least on Ubuntu 16.04 LTS. A quick google of `postgresql96-setup` suggests that this might only be the case for installs using `yum`.

What I can say is that using the configuration in the OP I was able to install PostgreSQL 9.6 on Ubuntu 16.04 LTS. I only left out the `"dynamic_shared_memory_type": "sysv"` portion after reading https://github.com/sous-chefs/postgresql/commit/0218940c29c91dc2cdc7d1fe91665dfb59b5d728.

damacus:
If anyone in this thread would like to support a PR for this, it would be great!

damacus:
Fixed in master, we now also support PG 10.

PR #406, opened 2017-01-06
Revert "Support for PostgreSQL 9.6 #370. not tested on RedHat." becau…

sauliusgrigaitis:
### Description

Revert "Support for PostgreSQL 9.6 #370. not tested on RedHat." because 'dynamic_shared_memory_type' needed only on some platforms/versions. I mostly run Ubuntu 16.04 LTS with PostgreSQL 9.6.1 now and this setting isn't needed any more (it was needed on older versions). Also, according to PostgreSQL documention best value should be set by default. 
 
### Issues Resolved

Do not set value that PostgreSQL should set by default.


damacus:
Looks good. 

Thanks.

PR #410, opened 2017-02-15
Kitchen refactor

AgarFu:
### Description

Removes the duplicity between .kitchen.yml and .kitchen.docker.yml.

There are only docker configs in .kitchen.docker.yml all the test suites definition, attributes, etc are in .kitchen.yml

I've included all the client suites/platforms in the travis file, do we really need to support centos-5?
### Contribution Check List

- [X] All tests pass.
- [X] New functionality has been documented in the README if applicable

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -24,10 +24,6 @@ platforms:
     driver:
       box: bento/debian-8.5
 
-  - name: centos-5
-    driver:
-      box: bento/centos-5.11
-
   - name: centos-6
     driver:
       box: bento/centos-6.8
@@ -39,7 +35,6 @@ platforms:
   - name: fedora-25
     driver:
       box: bento/fedora-25
-    run_list: yum::dnf_yum_compat
 
   - name: ubuntu-12.04
     driver:
@@ -92,7 +87,7 @@ suites:
   attributes:
     postgresql:
       password:
-        postgres: 'iloverandompasswordsbutthiswilldo'
+        postgres: iloverandompasswordsbutthiswilldo
 
 - name: apt-pgdg-server
   run_list:
@@ -103,17 +98,17 @@ suites:
     postgresql:
       enable_pgdg_apt: true
       version: '9.4'
-      dir: '/etc/postgresql/9.4/main'
+      dir: /etc/postgresql/9.4/main
       server:
-        service_name: 'postgresql'
+        service_name: postgresql
         packages: [ 'postgresql-9.4' ]
       client:
         packages: [ 'postgresql-client-9.4', 'libpq-dev' ]
       password:
-        postgres: 'iloverandompasswordsbutthiswilldo'
+        postgres: iloverandompasswordsbutthiswilldo
       config:
-        ssl_cert_file: '/etc/ssl/certs/ssl-cert-snakeoil.pem'
-        ssl_key_file: '/etc/ssl/private/ssl-cert-snakeoil.key'
+        ssl_cert_file: /etc/ssl/certs/ssl-cert-snakeoil.pem
+        ssl_key_file: /etc/ssl/private/ssl-cert-snakeoil.key
 
 - name: yum-pgdg-server
   run_list:
@@ -126,12 +121,12 @@ suites:
       version: '9.4'
       server:
         packages: [ 'postgresql94-server' ]
-        service_name: 'postgresql-9.4'
+        service_name: postgresql-9.4
       client:
         packages: [ 'postgresql94', 'postgresql94-devel' ]
-      setup_script: 'postgresql94-setup'
+      setup_script: postgresql94-setup
       password:
-        postgres: 'iloverandompasswordsbutthiswilldo'
+        postgres: iloverandompasswordsbutthiswilldo
 
 - name: apt-pgdg-client-ruby
   run_list:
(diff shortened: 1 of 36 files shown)


.kitchen.yml
```diff
@@ -1,21 +1,65 @@
+<%
+rpm_based_platforms = %w(centos-5 centos-6 centos-7 fedora-25)
+deb_based_platforms = %w(ubuntu-12.04 ubuntu-14.04 ubuntu-16.04 debian-7 debian-8)
+opensuse_platforms = %w(opensuse-13.2 opensuse-leap)
+%>
+---
 driver:
   name: vagrant
+  customize:
+    cableconnected1: 'on'
```
AgarFu: Yes it is causing problems testing debian-8 with vagrant, looks like it doesn't configure the nic if there is no link.

.kitchen.yml
```diff
@@ -1,21 +1,65 @@
...
+  name: inspec
+
 platforms:
-  - name: centos-6.8
-  - name: centos-7.2
-  - name: debian-7.11
-  - name: debian-8.6
+  - name: debian-7
+    driver:
+      box: bento/debian-7.11
```
AgarFu: My intention with this change is to name the platforms in a consistent way regardless the release version we use. Using generic names as debian-7 vs debian-7.11 we can define what release of debian we want to use via the box setting and we don't need to upgrade the list of exclusions from the test suites for instance and we can also use the same platform names in dokken, AWS, etc ... this is an extra level of abstraction but is useful to decouple the test suites definitions to the execution environment based on which kitchen driver you're using.

I can change it if you don't like it, but I really think it is more maintainable.

recipes/ruby.rb
```diff
@@ -31,7 +31,7 @@
   node.override['build-essential']['compile_time'] = true
   include_recipe 'build-essential'
 
-  if node['postgresql']['enable_pgdg_yum'] && platform_family?('rhel')
+  if node['postgresql']['enable_pgdg_yum'] && %w(rhel fedora).include?(node['platform_family'])
```
tas50: You can achieve this same thing with platform_family?('rhel', 'fedora')

.kitchen.yml
tas50: @AgarFu Is this an issue with the latest bento boxes? You have to manually update to pull down the new ones

.kitchen.yml
```diff
@@ -1,21 +1,65 @@
...
+
+  - name: debian-8
+    driver:
+      box: bento/debian-8.5
```
tas50: There's an 8.6 box out

.kitchen.yml
```diff
@@ -1,21 +1,65 @@
...
+    driver:
+      box: bento/centos-5.11
+
+  - name: centos-6
+    driver:
+      box: bento/centos-6.8
+
+  - name: centos-7
+    driver:
+      box: bento/centos-7.2
```
tas50: 7.3 is out too

.kitchen.yml
```diff
@@ -1,21 +1,65 @@
...
+
   - name: fedora-25
+    driver:
+      box: bento/fedora-25
     run_list: yum::dnf_yum_compat
```
tas50: Yep this isn't necessary with the new chef clients that include support for DNF out of the box

.kitchen.yml
```diff
@@ -1,21 +1,65 @@
...
-  - name: opensuse-13.2
-  - name: opensuse-leap-42.1
+
   - name: ubuntu-12.04
```
tas50: 12.04 goes EOL in April. We can probably yank testing now

.kitchen.yml
```diff
@@ -1,21 +1,65 @@
...
-  - name: debian-8.6
+  - name: debian-7
+    driver:
+      box: bento/debian-7.11
+
+  - name: debian-8
+    driver:
+      box: bento/debian-8.5
+
+  - name: centos-5
```
tas50: RHEL 5 goes out of standard support in April. It's probably safe to yank support now

.travis.yml
```diff
@@ -15,11 +15,16 @@ install: echo "skip bundle install"
...
+    - kitchen_refactor
 
 services: docker
 
 env:
+  global:
+  - KITCHEN_YAML=.kitchen.yml
+  - KITCHEN_LOCAL_YAML=.kitchen.docker.yml
   matrix:
+    #- INSTANCE=client-centos-5
```
tas50: This can just go away now. Big win there

.kitchen.yml
AgarFu: Seems that they are updated
```
* 'bento/debian-8.5' (v2.2.9) is up to date
* 'bento/debian-7.11' (v2.3.1) is up to date
```
What maybe is not updated is my virtual box

tas50:
@AgarFu Can you take a look at the tests that are failing. Some of them are failing on the kitchen suites not existing.

.kitchen.yml
tas50: Give it a try on the 8.6 image with the latest virtualbox/vagrant. There's been some issues in the recent past with cable connected and virtualbox

.travis.yml
```diff
@@ -15,10 +15,14 @@ install: echo "skip bundle install"
 branches:
   only:
     - master
+    - kitchen_refactor
```
tas50: I'll nuke this one post merge since we don't need it anymore

PR #413, opened 2017-02-19
Chef 13

damacus:
### Description
Turn on Chef 13 compat warnings


tas50:
This should get squashed

tas50:
Closing this out since it came in with my dokken fixes

PR #419, opened 2017-03-02
Test with Local Delivery and rename docker config to kitchen.dokken.yml

tas50:


tas50:
This fixes some Chef 13 issues too and turns on the deprecation warnings

.kitchen.dokken.yml
```diff
@@ -8,24 +8,9 @@ transport:
 
 provisioner:
   name: dokken
+  deprecations_as_errors: true
```
damacus: Doesn't this get inherited as part of .kitchen.yml?

test/cookbooks/test/recipes/ruby.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 file '/tmp/pg_gem_test.rb' do
   content "#!/opt/chef/embedded/bin/ruby\nrequire 'pg'"
-  mode 0755
+  mode '755'
```
damacus: Why? :/

tas50: Using strings here allow you to preserve any leading 0s. It's a rubocop-chef rule actually, which is what fixed this.

.kitchen.dokken.yml
tas50: That's a great question. I'll have to confirm that

PR #420, opened 2017-03-02
switch to realdirpath to resolve relative symlinks

teknofire:
### Description

This fixes an issue with detecting system timezone when using relative symlinks for `/etc/localtime`

On centos 7.2 `/etc/localtime` gets symlinked to `../usr/share/zoneinfo/UTC`

This would cause the following gsub to set `bestzonename` to `..UTC`
instead of just `UTC`

### Contribution Check List

- [x] All tests pass.
- [x] New functionality includes testing. (Nothing new)
- [x] New functionality has been documented in the README if applicable (Nothing new)

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -201,7 +201,7 @@ module Opscode
         # in spite of what the sysadmin had specified in the symlink.
         # (There are many duplicates under tzdir, with the same timezone
         # content appearing as an average of 2-3 different file names.)
-        path = ::File.readlink('/etc/localtime')
+        path = ::File.realdirpath('/etc/localtime')
         bestzonename = path.gsub("#{tzdir}/", '')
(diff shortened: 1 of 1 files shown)


libraries/default.rb
```diff
@@ -201,7 +201,7 @@ def scan_available_timezones(tzdir)
         # in spite of what the sysadmin had specified in the symlink.
         # (There are many duplicates under tzdir, with the same timezone
         # content appearing as an average of 2-3 different file names.)
-        path = ::File.readlink('/etc/localtime')
+        path = ::File.realdirpath('/etc/localtime')
```
josephholsten: TMYK [`File.realdirpath`](http://ruby-doc.org/core-2.4.0/File.html#method-c-realdirpath)

tas50:
👍

PR #432, opened 2017-07-12
Allow pgdg repository URL to be overridden with attributes

EricKnecht:
### Description

Currently the URL for the pgdg repository is hard coded. We have servers that are not allowed to access the public internet but can access our internal apt repository mirrors. We'd like to be able to override the URL and key for the pgdg repo to point at our internal mirrors. 

### Issues Resolved

None
### Contribution Check List

- [ ] All tests pass.


tas50:
Definitely a useful feature for airgapped environments.

PR #438, opened 2017-09-11
Postgresql Access resource

martinisoft:
### Description

This adds a new `postgresql_access` resource to manage the `pg_hba.conf` file instead of managing a pile of attributes to make this cookbook more reusable. This change also pushes the minimum chef requirement to 12.9+ to drop the need for `compat_resource`.

I've also provided a migration path with deprecation warnings and documented the resource along with the extension resource.

### Issues Resolved

None

### Contribution Check List

- [ ] All tests pass.


README.md
```diff
@@ -17,14 +17,41 @@ Installs and configures PostgreSQL as a client or a server.
 
 ### Chef
 
-- Chef 12.1+
+- Chef 12.5+
```
tas50: If we're doing custom resource it's actually 12.7. 12.5/12.6 had a pile of issues with action_class

README.md
```diff
@@ -17,14 +17,41 @@ Installs and configures PostgreSQL as a client or a server.
...
 
 ### Cookbooks
 
-- `compat_resource`
```
tas50: This was actually for apt_repository which we shipped in 12.9

martinisoft:
I have updated the initial comment. This will throw warnings for now in the test suite because it leverages the backward compatible method of attribute based access setup. We can update the test suite in the next release after the deprecation.

PR #448, opened 2017-10-24
Move methods into an action_class

damacus:
### Description

IIRC methods and includes should be in an action_class (like we do in ruby_rbenv)


tas50:
How did that work?

damacus:
Honestly I'm sure it didn't...

PR #454, opened 2017-10-24
Add a first pass for adding the official repo

damacus:
### Description

Add Repository custom resource & testing.

### Contribution Check List

- [x] All tests pass.

### DIFF
--- a/resources/repository.rb
+++ b/resources/repository.rb
@@ -1,0 +1,44 @@
+property :version, String, default: '10'
+property :minor_version, String, default: '10-1'
+
+action :add do
+  p_version = node['platform_version'].to_i
+  p_family = node['platform_family']
+  p_short = node['platform']
+
+  case node['platform_family']
+
+  when 'rhel'
+    remote_file "#{Chef::Config[:file_cache_path]}/pgdg.rpm" do
+      source "https://download.postgresql.org/pub/repos/yum/#{new_resource.version}/redhat/#{p_family}-#{p_version}-x86_64/pgdg-#{p_short}#{new_resource.version}-#{new_resource.minor_version}.noarch.rpm"
+      action :create
+    end
+
+    yum_package "#{Chef::Config[:file_cache_path]}/pgdg.rpm"
+  when 'debian'
+  #   apt_repository 'name' do
+  #    uri                   http://apt.postgresql.org/pub/repos/apt/ YOUR_UBUNTU_VERSION_HERE-pgdg main
+  #    distribution          String
+  #    components            Array
+  #    arch                  String
+  #    trusted               TrueClass, FalseClass
+  #    deb_src               TrueClass, FalseClass
+  #    keyserver             String
+  #    key                   String, Array
+  #    key_proxy             String
+  #    cookbook              String
+  #    cache_rebuild         TrueClass, FalseClass
+  #    sensitive             TrueClass, FalseClass
+  #   end
+  #
+  #   wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | \
+  #     sudo apt-key add -
+
+
+    apt_update 'update'
+  end
+end
+
+
+
+# stretch, jessie, wheezy,zesty
--- a/resources/repository.rb
+++ b/resources/repository.rb
@@ -17,7 +17,6 @@
 #
 
 property :version, String, default: '10'
-property :minor_version, String, default: '10-1'
 property :enable_pgdg, [true,false], default: false
 property :enable_pgdg_source, [true,false], default: false
 property :enable_pgdg_updates_testing, [true,false], default: true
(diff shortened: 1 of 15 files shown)


damacus:
Tests in this test suite pass. 

They're just testing the default and string interpolation work. 

Not adding documentation for this resource as I expect it'll only be used by other resources.

resources/repository.rb
```diff
@@ -0,0 +1,88 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+property :version, String, default: '10'
+property :minor_version, String, default: '10-1'
```
tas50: This isn't actually used

resources/repository.rb
```diff
@@ -0,0 +1,88 @@
...
+property :enable_pgdg_source_updates_testing, [true,false], default: false
+
+default_action :add
+
+action :add do
+  case node['platform_family']
+
+  when 'rhel'
+    remote_file "/etc/pki/rpm-gpg/RPM-GPG-KEY-PGDG-#{new_resource.version}" do
+      source "https://download.postgresql.org/pub/repos/yum/RPM-GPG-KEY-PGDG-#{new_resource.version}"
```
tas50: This should be a property with this value lazily evaluated

resources/repository.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+
+    yum_repository "PostgreSQL #{new_resource.version} $releasever - $basearch" do
```
tas50: It seems like instead of the 3 different properties we could make a property called yum_repos (or whatever) and pass an array. That array would create the various repos and default to the non-source repos.

resources/repository.rb
```diff
@@ -0,0 +1,88 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+property :version, String, default: '10'
+property :minor_version, String, default: '10-1'
+property :enable_pgdg, [true,false], default: false
```
tas50: pretty sure this was meant to default to true right?

resources/repository.rb
```diff
@@ -0,0 +1,88 @@
...
+      notifies :run, 'bash[apt-key-add]', :immediately
+    end
+
+    bash 'apt-key-add' do
+      code "sudo apt-key add #{Chef::Config[:file_cache_path]}/ACCC4CF8.asc"
+      action :nothing
+    end
+
+    apt_repository 'name' do
+     uri          "http://apt.postgresql.org/pub/repos/apt/"
```
tas50: https this. It'll break people without the apt https package, but that's ok

resources/repository.rb
```diff
@@ -0,0 +1,88 @@
...
+     components   ["main"]
+     distribution "#{node['lsb']['codename']}-pgdg"
+     cache_rebuild true
+    end
+
+    apt_update 'update'
```
tas50: This shouldn't be necessary as the apt_update forces a cache update on the repo

resources/repository.rb
damacus: I did in the end I'll push this up in a bit

resources/repository.rb
damacus: Awesome, ta.

resources/repository.rb
damacus: Done

resources/repository.rb
damacus: There's only one here? Unless I'm mistaken.

resources/repository.rb
damacus: Which bit sorry?

resources/repository.rb
```diff
@@ -0,0 +1,88 @@
...
+       gpgkey      "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-PGDG-#{new_resource.version}"
+    end
+
+  when 'debian'
+    remote_file "#{Chef::Config[:file_cache_path]}/ACCC4CF8.asc" do
+      source "https://www.postgresql.org/media/keys/ACCC4CF8.asc"
+      notifies :run, 'bash[apt-key-add]', :immediately
+    end
+
+    apt_update 'update'
```
damacus: Without this here, we don't find an upto date mirror for apt-transport so the install fails

resources/repository.rb
tas50: The location of the GPG value. Anyone in an airgapped environment is going to want to point this whole thing at internal mirrors.

resources/repository.rb
```diff
@@ -0,0 +1,88 @@
...
+    package 'apt-transport-https'
+
+    bash 'apt-key-add' do
```
tas50: apt_repository can handle pulling this key in. Check the resource docs page

resources/repository.rb
damacus: Good call

resources/repository.rb
damacus: Yeah. I saw this, but it's a 13.4 feature. 

```
If a keyserver is provided, this is assumed to be the fingerprint; otherwise it can be either the URI of GPG key for the repo, or a cookbook_file. Default value: nil.

New in Chef client 13.4.
```

tas50:
Merging this to open a PR to tweak it further.

PR #461, opened 2017-10-30
Deprecate the legacy recipes

tas50:


BrandonWLee:
Hey @damacus, why are these recipes now deprecated?

damacus:
Hey @BrandonWLee 

as this is an application cookbook, recipes are not the best way to provide functionality out of the box. 

What we're doing here is writing something close to Core Chef resources. take for example the user resource, and providing a recipe for that, wouldn't be very flexible. 

What is provided, but not currently documented, are example use cases in the form of test suites. 

See this directory for usage: https://github.com/sous-chefs/postgresql/tree/master/test/cookbooks/test/recipes

BrandonWLee:
I see! I noticed the config_pgtune recipe was removed. Will that eventually be incorporated into a resource or is that going to be removed for good?

damacus:
We'll probably add a resource for it eventually.

If you/we can distill it down into something that we can tackle I don't see why not! 

Fancy having a crack at it?

BrandonWLee:
I'll see if I have time for it :)
Thank you for your responses @damacus.

PR #464, opened 2017-11-01
Initial work to support Fedora / Amazon in server resource

tas50:
This is a phase 1 of getting the server install to work on multiple platforms


tas50:
@damacus This would pass except for a bug in inspec that I'm working out. It's good to merge and fixes a bunch of issues with the current server setup

PR #470, opened 2017-12-12
More cleanup

damacus:
### Description

[Describe what this change achieves]

### Issues Resolved

[List any existing issues this PR resolves]

### Contribution Check List

- [ ] All tests pass.


resources/pg_gem.rb
```diff
@@ -1,7 +1,8 @@
-# frozen_string_literal: false
+# frozen_string_literal: true
+# encoding: utf-8
```
tas50: Do we need to define the encoding here?

resources/pg_gem.rb
```diff
@@ -47,76 +58,39 @@
...
-      lib_builder.run_action(:run)
-
-      lib_maker = execute 'make pg gem lib' do
-        command 'make'
-        cwd ext_dir
-        action :nothing
-      end
-      lib_maker.run_action(:run)
+  if ruby_version < 2.0
+    Chef::Log.fatal("pg gem requires a system Ruby installation of 2.0 or higher. \n Please install a global Ruby")
```
tas50: You want to raise here so you actually fail the run

resources/pg_gem.rb
```diff
@@ -47,76 +58,39 @@
...
+  gem_package 'pg' do
+    clear_sources new_resource.clear_sources if new_resource.clear_sources
+    include_default_source new_resource.include_default_source if new_resource.include_default_source
+    gem_binary new_resource.gem_binary if new_resource.gem_binary
+    options new_resource.options if new_resource.options
+    source new_resource.source if new_resource.source
+    timeout new_resource.timeout if new_resource.timeout
+    version new_resource.version if new_resource.version
+    action :install
+    not_if { ruby_version < 2.0 }
```
tas50: We shouldn't need to double guard against old ruby

resources/pg_gem.rb
damacus: Cookstyle probably.

resources/pg_gem.rb
tas50: I meant you should just raise "stuff be wrong!" instead of the chef log.

resources/pg_gem.rb
```diff
@@ -47,76 +58,36 @@
...
-      end
-      lib_builder.run_action(:run)
-
-      lib_maker = execute 'make pg gem lib' do
-        command 'make'
-        cwd ext_dir
-        action :nothing
-      end
-      lib_maker.run_action(:run)
+  raise ArgumentError, "pg gem requires a system Ruby installation of 2.0 or higher." if uby_version < 2.0
```
duncaan: Did you mean `ruby_version`?

damacus:
Most of ruby_gem_pg is a mess right now.

There's a few discussions about why. So I'm not surprised if there are some
bugs in here.

We/i really want this to work on all platforms or none.

PR #475, opened 2017-12-28
Add back Amazon Linux testing to Travis

tas50:


damacus:
🤔 should we also fix Amazon as part of this PR, rather than breaking the tests?

Issue #466, opened 2017-11-20
Where to add pg_hba.conf lines


WojTecH94:
Hi! I am quite new to chef and trying to setup postgresql installation, from this git repo.

### Scenario:
Add a lines to pg_hba.conf

### Steps to Reproduce:
Trying to add:

> postgresql_access 'local_postgres_superuser' do
>   comment 'Local postgres superuser access'
>   access_type 'local'
>   access_db 'all'
>   access_user 'postgres'
>   access_addr nil
>   access_method 'ident'
> end

like in description of cookbook but don't know to which file and which place should I add it?

### Expected Result:
A lines:

```
 # Local postgres superuser access
 local   all             postgres                                ident
```

should be added to pg_hba.conf

### Actual Result:
error while running chef-client on node:
> undefined method `postgresql_access'

damacus:
Welcome @WojTecH94 ! It might be worth dropping into #sous-chef on the [chef community slack](https://www.google.co.uk/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwj6rbyTnq3YAhWnKcAKHRJ7B30QFggpMAA&url=http%3A%2F%2Fcommunity-slack.chef.io%2F&usg=AOvVaw0dWNaK4RejWCaB5vjZu_Is) for a walk through of how to use the cookbook if you're still stuck!

damacus:
Closing due to inactivity. 

Hi @WojTecH94 if you're still having this issue after the next major release of the cookbook, please reopen! 

Thanks, 
Dan

Issue #426, opened 2017-03-27
server_debian does not correctly handle alternate data_directory settings


AltxSF:
### Cookbook version
6.1.1
### Chef-client version
12.18.3
### Platform Details
ubuntu 16.04
### Scenario:
Install postgres with non-default data_directory
### Steps to Reproduce:
set data_directory to something other than the default
node.override[:postgresql][:config][:data_directory] = '/postgres/data'

I believe the problem is the following snippet from recipe/server_debian.rb

```
execute 'Set locale and Create cluster' do
  command 'export LC_ALL=C; /usr/bin/pg_createcluster --start ' + node['postgresql']['version'] + ' main'
  action :run
  not_if { ::File.directory?('/etc/postgresql/' + node['postgresql']['version'] + '/main') }
end
```

The command fails to include the --datadir argument, the end result of this is that while postgresql.conf refers to the non-standard data_directory, this directory remains empty as pg_createcluster creates the various data files in its default data directory location (/var/lib/postgresql/9.5/main).

adding "--datadir" + node['postgreql]['config']['data_directory'] to the end of the command would solve this problem.

shortdudey123:
edit: fixed formatting of chef resource code block

damacus:
Hey @AltxSF sorry for the long triage on this one!

Can you try the new custom resource `postgresql_server_install`? 

Right now it's not configurable, but with a little tweaking, it could be fixed to do be.

damacus:
This should now be fixed as part of #481 and is now on master. 

Please reopen if this isn't the case!

Issue #403, opened 2017-01-04
This recipe is not creating a postgresql.conf 


mreith101:
### Cookbook version
6.0.0

### Chef-client version

12.14.89

### Platform Details
6.5

### Scenario:

I'm trying to install PostgreSQL into a custom folder /var/lib/pgsql/9.5

### Steps to Reproduce:

Running chef-client and including: 

include_recipe "postgresql::server" (in a wrapper cookbook) is installing ok but isn't creating a postgresql.conf

node.override['postgresql']['config']['listen_addresses'] = 'fake1local' an attribute override that I set for a test case.

### Expected Result:

Exepect postgresql.conf of any form (preferably with my custom changes) in /var/lib/pgsql/9.5

### Actual Result:

PostgreSQL fails to start because the configuration file isn't being generated properly.

damacus:
Fixed by #481 .

Issue #329, opened 2016-01-26
Bad generated config (9.4)


SerjoPepper:
Hi. I use version `"version": "3.4.24"` with follow config:

```
override['postgresql']['config_pgtune']['db_type'] = 'web'
override['postgresql']['password']['postgres'] = 'pass'

override['postgresql']['version'] = "9.4"
override['postgresql']['dir'] = "/etc/postgresql/9.4/main"
override['postgresql']['server']['service_name'] = "postgresql"
override['postgresql']['client']['packages'] = ["postgresql-client-9.4","libpq-dev"]
override['postgresql']['server']['packages'] = ["postgresql-9.4"]
override['postgresql']['contrib']['packages'] = ["postgresql-contrib-9.4"]
override['postgresql']['contrib']['extensions'] = [
  "pageinspect",
  "pg_buffercache",
  "pg_freespacemap",
  "pgrowlocks",
  "pg_stat_statements",
  "pgstattuple",
  "btree_gin"
]
```

Invoke recipes:

```
include_recipe 'postgresql::contrib'
include_recipe 'postgresql::server_debian'
include_recipe 'postgresql::client'
include_recipe 'postgresql::config_pgtune'
```

All was installed OK, but I have bad config with references to 9.3 version

```
vagrant@vagrant-ubuntu-trusty-64:~$ sudo cat /etc/postgresql/9.4/main/postgresql.conf
# PostgreSQL configuration file
# This file was automatically generated and dropped off by chef!
# Please refer to the PostgreSQL documentation for details on
# configuration settings.

checkpoint_completion_target = '0.7'
checkpoint_segments = 8
data_directory = '/var/lib/postgresql/9.3/main'
datestyle = 'iso, mdy'
default_statistics_target = 100
default_text_search_config = 'pg_catalog.english'
effective_cache_size = '2GB'
external_pid_file = '/var/run/postgresql/9.3-main.pid'
hba_file = '/etc/postgresql/9.3/main/pg_hba.conf'
ident_file = '/etc/postgresql/9.3/main/pg_ident.conf'
listen_addresses = 'localhost'
log_line_prefix = '%t '
maintenance_work_mem = '176MB'
max_connections = 200
port = 5432
shared_buffers = '704MB'
ssl = on
ssl_cert_file = '/etc/ssl/certs/ssl-cert-snakeoil.pem'
ssl_key_file = '/etc/ssl/private/ssl-cert-snakeoil.key'
[...]

PR #481, opened 2018-01-12
fix postgresql configuration and access resources

### LINKED ISSUE #472: PostgreSQL Config Resource
## Feature

As a postgreSQL admin
I want my postgreSQL server configured the same reliably
And no matter which platform I chose listed in the supported platforms it should work
So that I can have some level of configuration disaster recovery

## Supporting Information

None yet. 

## Related Issues

- #310 
- #329

YpNo:
### Description

Review master version which fix postgresql configuration and access resource

### Issues Resolved

#472
#329
#310 
#466
Maybe #426
#403

### Contribution Check List

- [x] Tested on Debian7, Debian8 and CentOS7


libraries/resource_helpers.rb
```diff
@@ -22,10 +22,27 @@ module Helpers
...
-        "/var/lib/postgresql/#{new_resource.version}/main"
+        "/var/lib/postgresql/#{node.run_state['postgresql']['version']}/main"
       end
     end
+
+    def conf_dir
+      case node['platform_family']
+      when 'rhel', 'fedora', 'amazon'
+        "/var/lib/pgsql/#{node.run_state['postgresql']['version']}/data"
+      when 'debian', 'ubuntu'
```
tas50: no need for ubuntu here. Debian platform family includes ubuntu

resources/access.rb
```diff
@@ -28,26 +28,25 @@
...
 action :grant do
   with_run_context :root do # ~FC037
-    edit_resource(:template, "#{postgresql_data_dir}/pg_hba.conf") do |new_resource|
+    find_resource(:service, 'postgresql')
+    edit_resource(:template, "#{conf_dir}/pg_hba.conf") do |new_resource|
       source new_resource.source
       cookbook new_resource.cookbook
       owner 'postgres'
       group 'postgres'
-      mode '0600'
```
tas50: This actually needs to be a string so we don't throw out the leading 0

libraries/resource_helpers.rb
YpNo: Agree

resources/access.rb
YpNo: I didn't know. I though it was a good practice to not use string for mode parameter

resources/repository.rb
```diff
@@ -69,23 +69,13 @@
...
-    bash 'apt-key-add' do
-      code "sudo apt-key add #{Chef::Config[:file_cache_path]}/ACCC4CF8.asc"
-      action :nothing
-    end
-
     apt_repository 'postgresql_org_repository' do
       uri          'https://download.postgresql.org/pub/repos/apt/'
       components   ['main', new_resource.version.to_s]
       distribution "#{node['lsb']['codename']}-pgdg"
+      key new_resource.apt_gpg_key_uri
```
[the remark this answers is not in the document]
YpNo: Wrong, I have tested with 12.14.60 version that is the requirement in README :)
But delayed_action used in cookbook depends on 12.16.+ version. I will modify the README

YpNo:
reviewed comment Fixed ✅

YpNo:
So, I don't understand why the build fail on Debian7 and Ubuntu1404. The service seems not running at the end...

damacus:
I'm going to pull this in straight before mine.

PR #485, opened 2018-01-18
Update README and check if node is slave

YpNo:
### Description

Update README.md, do a check if node is a PostgreSQL slave to not execute query on it, and fix service management

### Contribution Check List

- [x] Test on Debian 7 and CentOS6


resources/database.rb
```diff
@@ -43,7 +38,7 @@
   bash "Create Database #{new_resource.database}" do
     code createdb
     user new_resource.user
-    not_if { database_exists?(new_resource) }
+    not_if { is_slave? && database_exists?(new_resource) }
```
damacus: This should probably be an OR here.

resources/extension.rb
```diff
@@ -27,7 +27,7 @@
     code psql("CREATE EXTENSION IF NOT EXISTS \"#{new_resource.extension}\"")
     user 'postgres'
     action :run
-    not_if { extension_installed? }
+    not_if { is_slave? && extension_installed? }
```
damacus: OR again?

resources/user.rb
```diff
@@ -23,13 +23,14 @@
...
 property :encrypted_password, String
+property :valid_until,        String
 
 action :create do
   execute "create postgresql user #{new_resource.name}" do # ~FC009
     user 'postgres'
     command %(psql -c "CREATE ROLE #{role_sql(new_resource)}")
     sensitive true
-    not_if { user_exists?(new_resource) }
+    not_if { is_slave? && user_exists?(new_resource) }
```
damacus: OR*?

resources/database.rb
YpNo: good point :)

YpNo:
I have crate un PR in the locale cookbook to fix the builds: https://github.com/chef-cookbooks/locale/pull/29

damacus:
LGTM, I don't want to block this work on the locale stuff.

PR #487, opened 2018-01-19
Update readme

YpNo:
### Description

Update README + Fix service scope context + default template in database resource

### Contribution Check List

- [x] Tested on Debian 7 and CentOS 7

### DIFF
--- a/libraries/resource_helpers.rb
+++ b/libraries/resource_helpers.rb
@@ -43,8 +43,6 @@ module PostgresqlCookbook
     end
 
     def postgresql_service
-      resources(service: 'postgresql')
-    rescue Chef::Exceptions::ResourceNotFound
       service 'postgresql' do
         service_name lazy { platform_service_name }
         supports restart: true, status: true, reload: true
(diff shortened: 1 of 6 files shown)


libraries/resource_helpers.rb
```diff
@@ -42,6 +42,16 @@ def platform_service_name
       platform_family?('rhel', 'amazon', 'fedora') ? "postgresql-#{node.run_state['postgresql']['version']}" : 'postgresql'
     end
 
+    def postgresql_service
+      resources(service: 'postgresql')
```
damacus: Can't we just use `platform_service_name`? rather than failing out?

YpNo: I don't understand what you want me to do / what you are saying ?
I have used this trick (inspired by andy on slack) because when you declare a service resource in custom resource and you notify it in other resource, the other resource don't find the service.
May be there is better to do that, and I want to know !

damacus:
Wondering whether you can just get the service name from the method? Rather
than trying the default then falling back to the method?

YpNo:
Updated.

YpNo:
Just FI, just call resources(service: 'postgresql') in method don't fix the issue.

damacus:
Awesome! Sorry this took so long to merge in!

PR #488, opened 2018-01-20
Database resource should ignore blank template

gregmolnar:
The default template for the resource is a blank string and if it isn't overridden, it generates a syntax error in the query.


damacus:
Fixed in #491 as we had some README changes, thanks for the fix!

PR #501, opened 2018-01-28
Fix postgresql_access resource

scopenco:
### Description

`postgresql_access` is broken. No one resource manages pg_hba.conf except `postgresql_access` but it uses `edit_resource` to change nonexisting resource `template`. 

Please correct me where i'm wrong. Thanks.

### DIFF
--- a/resources/access.rb
+++ b/resources/access.rb
@@ -24,30 +24,24 @@ property :access_db,     String, required: true, default: 'all'
 property :access_user,   String, required: true, default: 'postgres'
 property :access_addr,   String, required: true
 property :access_method, String, required: true, default: 'ident'
-property :notification,  Symbol, required: true, default: :reload
 
 action :grant do
-  with_run_context :root do # ~FC037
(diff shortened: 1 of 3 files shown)


damacus:
Awesome @scopenco ! 

Could you throw in a config file test for this please?

martinisoft:
I'm not sure how this is broken? Got a failing test? This is using the accumulator pattern which is still being fleshed out in Chef. I [wrote about this](https://blog.dnsimple.com/2017/10/chef-accumulators/) to explain the pattern so I'd love to see a failing case along with which Chef version you were running to hit this issue.

damacus:
Closing this down as it doesn't look broken right now. 

Unless we can write some tests to prove this hypothesis wrong, this fixes a non-problem.

martinisoft:
It's worth noting that Chef 13.7 [has a regression around node attributes](https://github.com/chef/chef/issues/6792) which completely screws up this pattern and is currently being addressed by Chef.

damacus:
That bug affects haproxy and samba in that case too. But there's no point adding information to that ticket. Unless we can be constructive(!)

PR #503, opened 2018-02-02
Fixup extension SQL execution

martinisoft:
### Description

This PR fixes a bug in the extension resource where there were conflicting psql methods so I did several things here:

* Renamed the method to be more clear of its purpose
* Marked the old method name for later removal
* Removed the duplicate method in the resource
* Added testing to the helpers via rspec and as much plain 'ol ruby as possible

### Issues Resolved

None

### Contribution Check List

- [x] All tests pass.

### DIFF
--- a/libraries/resource_helpers.rb
+++ b/libraries/resource_helpers.rb
@@ -38,8 +38,12 @@ module PostgresqlCookbook
     end
 
     # determine the platform specific service name
-    def platform_service_name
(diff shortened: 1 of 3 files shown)


libraries/resource_helpers.rb
```diff
@@ -19,27 +19,31 @@
 
 module PostgresqlCookbook
   module Helpers
-    def data_dir
+    def data_dir(version = node.run_state['postgresql']['version'])
```
damacus: So much tidier. It's like you program ruby regularly! :D

damacus:
Leaving all those comments in the git log, because super useful. thanks! <3

PR #505, opened 2018-02-02
Cleanup and separate server install actions

martinisoft:
### Description

This PR separates the install action of the server_install resource into install and create to separate concerns and simplify things. I would like to ask @damacus and @tas50 if they agree with me wanting to do the following in this PR:

* Rename `server_install` to `server` so we get `postgresql_server` as a resource and `:install` as an action which semantically makes a lot more sense.
* Remove the client install from `server_install`
* Move or remove the database initialization code. If you want to keep it, maybe a `cluster` resource since that is the terminology postgres tooling uses (ever use clusterctl?)
* Push a few more action_class methods into helpers for easier testing (I started some tests in #503 )

I think ideally, this cookbook should help you install a postgres server and configure it in an elegant way and in a sort of future-proofy way if possible so we're not hard coding all the options. It should also ditch the notion of a global PG version as you can _technically_ have multiple copies and clusters of postgres installed. You could install a 9.6 cluster and a 10.0 cluster then setup replication between the two via `pg_logical`, but as this cookbook is currently written, that won't happen.

Right now these changes in the PR get 3 of the 9 platforms passing for server-install and the other 6 are getting issues caused by the config resource so I can address those in a separate PR. I still need to fix the `setup_repo` logic since it is missing here, but forced into the client resource.

### Issues Resolved

None

### Contribution Check List

- [ ] All tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,11 +124,12 @@ end
 
 ### postgresql_server_install
 
-This resource install PostgreSQL client and server packages.
+This resource installs the PostgreSQL server
 
 #### Actions
 
 - `install` - (default) Install client and server packages
+- `create` - Initialize the database
 
 #### Properties
 
@@ -148,8 +149,13 @@ Name                | Types           | Description                             
 To install PostgreSQL server, set you own postgres password and set another service port.
 ```
 postgresql_server_install 'My Postgresql Server install' do
+  action :install
+end
+
+postgresql_server_install 'Setup my postgresql 9.5 server' do
   password 'MyP4ssw0d'
   port 5433
+  action :create
 end
 ```
 
--- a/resources/server_install.rb
+++ b/resources/server_install.rb
@@ -17,7 +17,6 @@
 #
 
 property :version,           String, default: '9.6'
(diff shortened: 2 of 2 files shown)


damacus:
I think this makes sense. 

This is a general pattern across all the DB cookbooks that we need to decide on really. 

a)
-  Install software + init

b) 
- Install software
- Create database


The second option is more work for the user and would require more documentation, *but* makes more sense.
Also on debian platforms the two are mangled together. :shakes-fist:

I'm really on the fence with this one, seeing as it only applies to RHEL based platforms.

damacus:
@martinisoft as we discussed on Slack. 

init_db resource
install resource

It might be worth starting this one over?

martinisoft:
> @martinisoft as we discussed on Slack.
> 
> init_db resource
> install resource
> 
> It might be worth starting this one over?

I think this PR is worth saving. Will work on moving that action out to a separate resource and getting tests around it.

damacus:
OK merging this in so we can separate this out after

PR #511, opened 2018-03-13
Six fixups

dschlenk:
### Description

Get rid of deprecation warnings in the extension resource. Also includes changes to make the test kitchen suites all pass. Based on the 6.1.1 release as it looks like a lot of other changes are in the works in master. Feel free to reject - my thought was maybe use something like this to release a 6.1.2 while whatever the plan for master gets finished, particularly with Chef client 14 coming soon. If so this should apply cleanly to a branch made from the 6.1.1 release.

### Contribution Check List

- [X] All tests pass.
- [NA] New functionality includes testing.
- [NA] New functionality has been documented in the README if applicable

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -72,6 +72,8 @@ suites:
   run_list:
   - recipe[postgresql::ruby]
   - recipe[test::ruby]
+  excludes:
+    - centos-6
 
 - name: server
   run_list:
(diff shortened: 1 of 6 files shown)


damacus:
Oh yeah sorry we're going to have to reject this see #512 for a tonne of information on the current status. 

We're really close the biggest hurdle is removing the PG-gem because it's a bag of cats and never really worked!

Issue #514, opened 2018-04-07
postgresql_database idempotency


Plork:
### Cookbook version
6.1.1 from 'master' commit: d476cdf

### Chef-client version
13.8.5

### Platform Details
Centos-7.3 (bento)

### Scenario:
Creating a Database/ and User as owner.

#### attributes.rb
```ruby
default['testdatabase']['db_name'] = 'database'
default['testdatabase']['db_username'] = 'user'
default['testdatabase']['db_password'] = 'password'
```

#### recipe.rb
```ruby
postgresql_server_install 'PostgreSQL Package' do
  version '9.6'
  action %i(install create)
end

postgresql_user 'PostgreSQL User' do
  action :create
  user node['testdatabase']['db_username']
  password node['testdatabase']['db_password']
end

postgresql_database 'PostgreSQL Database' do
  database node['testdatabase']['db_name']
  action :create
  owner node['testdatabase']['db_username']
end
```

### Actual Result:
Running the recipe the second time gives
` STDERR: createdb: database creation failed: ERROR:  database "database" already exists`

The first time the recipe runs fine. Seems the "database_exists?" doesn't work.
Running the command manually gives:

```
-bash-4.2$ psql -c "SELECT datname from pg_database WHERE datname='database'" postgres | grep database
 database
```

Plork:
Adding some logging in the `:create` action gives:
```
* log[false] action write
* bash[Create Database database] action run
```

```ruby
  log database_exists?(new_resource)

  bash "Create Database #{new_resource.database}" do
    code createdb
    user new_resource.user
    not_if { database_exists?(new_resource) }
  end
```

damacus:
Fixed by #515

PR #515, opened 2018-04-07
the name of postgresql_database is database

Plork:
### Description

change new_resource.name to new_resource.database for the database_exists helper.

### Issues Resolved

https://github.com/sous-chefs/postgresql/issues/514

### DIFF
--- a/resources/database.rb
+++ b/resources/database.rb
@@ -63,15 +63,16 @@ action_class do
   include PostgresqlCookbook::Helpers
 
   def database_exists?(new_resource)
-    sql = %(SELECT datname from pg_database WHERE datname='#{new_resource.name}')
+    sql = %(SELECT datname from pg_database WHERE datname='#{new_resource.database}')
 
-    exists = %(psql -c "#{sql}" postgres)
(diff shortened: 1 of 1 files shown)


resources/database.rb
```diff
@@ -63,15 +63,16 @@
...
+    exists << " -U #{new_resource.user}" if new_resource.user
     exists << " --host #{new_resource.host}" if new_resource.host
     exists << " --port #{new_resource.port}" if new_resource.port
-    exists << " | grep #{new_resource.name}"
+    exists << " | grep #{new_resource.database}"
 
     cmd = Mixlib::ShellOut.new(exists, user: 'postgres')
     cmd.run_command
-    cmd.exitstatus == 0
+    cmd.exitstatus.zero?
```
damacus: Cookstyle doesn't like this, even though I prefer the style (and use this in other Ruby projects!)

damacus:
Thanks! Nice find. 

As discussed we should/will turn on the multiple converge to pick up bugs like it not finding the database...!

Issue #500, opened 2018-01-24
Functional Test: Ident resource


damacus:

damacus:
Closed by #517

Issue #512, opened 2018-03-23
Feedback: Major breaking changes upcoming, need some feedback


martinisoft:
Hey everyone! I'm in the midst of doing a fair bit of reconstruction work on the cookbook which is slated for the 7.x release. The things it will do are:

* Chef 13.8+ compatibility. Since Chef 14 comes out in April and Chef 12 will be EOL at the end of April, it is time to move everything up.
* Drastically alter the way the cookbook works under the covers. For the longest time it was dependent on the pg gem which was no picnic to install into the Chef Omnibus and thus it has temporarily been removed from this upcoming release until there is a cleaner/better/testable solution. Now it is a series of shell outs to the postgres toolchain to directly manipulate systems.
* Re-factors the very extensive attributes away from the cookbook entirely. For a very long time this cookbook used the library cookbook anti-pattern that is attributes. Since you can technically have multiple versions of postgresql installed on a given system, the current way it is setup does not allow for you to manage all of those clusters. Going forward, the first (small) step to supporting clusters is to specify a version number. Not only does this correspond to package installs, but it also refers to the cluster name as well which on _most_ systems is named after the `PG_VERSION` that is present in those folders. A future refactor can take the postgres cluster scheme into better account. If you have a solid use case for this too, please open an issue so we can examine it and see how we can update the resources to account for it.
* Creates a lot more testing to verify correctness. There are now integration tests to cover all of the resources sometimes multiple times. This will slow down the overall test suite speed, but I'll trade test coverage for CI speed every day.
* Simplifies the library helpers and removes them from the global namespace. Forcing the postgres global helpers upon everyone is not a very nice thing to do so there will be examples and documentation on how to do this better in your wrapper cookbook.

Whew, yes this is one hell of a rewrite that needed to happen for the future maintainability of this cookbook. So, what I'd also like feedback on is how we roll this out to everyone using this cookbook. My idea for now is to cut a branch off the most recent 6 series release called 6.x then make a new patch version which adds a lot of `Chef::Log.warn`'s everywhere and some lengthy deprecation and migration documentation in the README. We can also leverage this branch to backport any more changes to make it Chef 13 compatible to buy users a bit more time to migrate up to the newer cookbook.

damacus:
@florian-asche can you try again using the current master branch? 

Should be good to go now.

florian-asche:
```
 * postgresql_user[fhem] action create
    * execute[create postgresql user fhem] action run
      
      ================================================================================
[...]

Issue #498, opened 2018-01-24
Functional Test: extension


damacus:

damacus:
Closed by #517

Issue #497, opened 2018-01-24
Functional Test: user


damacus:

YpNo:
Take it

YpNo:
Oh, @martinisoft has already work on it in his last PR. Great :)

damacus:
Closed by #517

PR #517, opened 2018-04-11
Library cleanup re-factor

martinisoft:
### Description

This is in relation to #512 where I am extracting all of the `action_class` helpers into a common helper library for more reliable unit testing. It also separates the resource notifications from the config resource [due to a bug in chef with delayed_action](https://github.com/chef/chef/issues/6987) so you'll now have more control over signaling. The other goal is to fully stabilize the unit and integration tests of the resources via the test cookbook which will resolve a whole host of other issues currently happening in master.

Right now this is still work in progress and I will summon a review when complete.

### Issues Resolved

- #496 
- #497 
- #498 
- #499 
- #500 

### Contribution Check List

- [ ] All tests pass.


resources/access.rb
```diff
@@ -24,18 +24,18 @@
...
   with_run_context :root do # ~FC037
     edit_resource(:template, "#{conf_dir}/pg_hba.conf") do |new_resource|
       source new_resource.source
       cookbook new_resource.cookbook
       owner 'postgres'
       group 'postgres'
       mode '0600'
-      variables['pg_hba'] ||= {}
-      variables['pg_hba'][new_resource.name] = {
+      variables[:pg_hba] ||= {}
```
damacus: Why symbols instead?

libraries/helpers.rb
```diff
@@ -302,6 +301,52 @@ def execute_sql(query, db_name)
...
+      end
+    end
+
+    def psql_command_string(database, query)
+      "psql -d #{database} <<< '\\set ON_ERROR_STOP on\n#{query};'"
+    end
+
+    # XXX: Remove me after removing this method elsewhere
+    def psql(database, query)
+      psql_command_string(database, query)
```
damacus: Thus, I think, can be removed now

resources/access.rb
```diff
@@ -24,18 +24,18 @@
 property :access_user,   String, required: true, default: 'postgres'
 property :access_addr,   [String, nil], default: nil
 property :access_method, String, required: true, default: 'ident'
-property :notification,  Symbol, required: true, default: :reload
 
 action :grant do
+  config_resource = new_resource
```
damacus: What does this bit do? Just provide a sensible alias?

resources/access.rb
```diff
@@ -24,18 +24,18 @@
 property :access_user,   String, required: true, default: 'postgres'
 property :access_addr,   [String, nil], default: nil
 property :access_method, String, required: true, default: 'ident'
-property :notification,  Symbol, required: true, default: :reload
```
YpNo: Why don't let the user the possibilty to choose trigger method for the service ? May be, I don't want to restart postgresql service (in cluster env) when I add/remove an user.

resources/access.rb
martinisoft: I prefer them because memory usage is better. Minor optimization I know, but I have the opportunity and I wanted to take it. As far as I know, other precedent hasn't been set in the released code for this.

resources/access.rb
martinisoft: So this has to do with context of execution because the `new_resource` yielded in the `edit_resource` block is not the same as the `new_resource` provided by the config resource therefore I memoize it so I can signal the parent resource.

So for example, if I attach a notification to a `postgresql_config` resource and I use the `new_resource` given inside the `edit_resource` block, it won't trigger any notifications until much later because it will have a different copy of the resource versus the original definition of `postgresql_config`.

libraries/helpers.rb
martinisoft: Yup, I have removed this. Good catch 👍

martinisoft:
Alright I now have the new ident resource covered by a full integration test so that resolves #500

Still have to work on the test failures around server install tests, but I am making great progress so far.

resources/access.rb
YpNo: If I understand just call notification in postgresql_access resource call like we do with chef resources ?
I agree but I have a brain conflict by thinking that I will need to notify each time I call the resource 🤣. If you have i.e 50 entries in your hba file, there will be 50 notifications too.

martinisoft: There won't be in terms of service notifications. Chef compacts them at the end of the run unless they are flagged as `:immediate` because by default all notifications are `:delayed` in resources.

resources/access.rb
```diff
@@ -45,11 +45,15 @@
...
       action :nothing
       delayed_action :create
-      notifies new_resource.notification, postgresql_service
+      notifies :trigger, config_resource, :immediately
     end
   end
 end
 
+action :trigger do
+  new_resource.updated_by_last_action(true)
```
damacus: This (and the ident resource) is throwing a FC085 warning. Do we either ignore this error or find another way of doing this, and I don't know how to in this particular case(!0

damacus:
A lot of the `server_install` bugs get fixed by #521 which is based on this branch. Give me a shout when you've pulled that in. 

We could do with more unit tests as you've pointed out. But if you could throw some example tests in with this branch we can farm out the tests. 

If we get this merged into master, we can iterate on it in there.

damacus:
I'd like to get this stage merged in now, and look at fixing the other tests in another pr. How's that sound?

martinisoft:
Formally requesting review since we've got green across the board. I'm going to break out the other re-factors into separate PRs but we're pretty close to a 7.0 release now.

libraries/helpers.rb
```diff
@@ -0,0 +1,133 @@
...
+    def execute_sql(query, db_name)
+      # query could be a String or an Array of String
+      statement = query.is_a?(String) ? query : query.join("\n")
+      cmd = shell_out("psql -q --tuples-only --no-align -d #{db_name} -f -",
+                      user: 'postgres',
+                      input: statement)
+      # If psql fails, generally the postgresql service is down.
+      # Instead of aborting chef with a fatal error, let's just
+      # pass these non-zero exitstatus back as empty cmd.stdout.
+      if cmd.exitstatus == 0 && !cmd.stderr.empty?
```
tas50: I think you can do !cmd.error? for the 2nd part here

libraries/helpers.rb
```diff
@@ -0,0 +1,133 @@
...
+      when 'rhel', 'fedora', 'amazon'
+        "/var/lib/pgsql/#{version}/data"
+      when 'debian'
+        "/etc/postgresql/#{version}/main"
+      end
+    end
+
+    # determine the platform specific service name
+    def platform_service_name(version = node.run_state['postgresql']['version'])
+      if %w(rhel amazon fedora).include?(node['platform_family'])
```
tas50: platform_family?('fedora', 'amazon', 'fedora') is your friend

martinisoft: Until you have to unit test it, then you're crying in a corner.

martinisoft: I love helpers like that, but without a stable, sanctioned mock of it, I'll build my own.

PR #521, opened 2018-04-16
Damacus/patch libraries

damacus:
### Description

Get server_install working. 

### Issues Resolved

Server install for centos-7 now works locally.
### Contribution Check List

- [ ] All tests pass.

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -5,10 +5,10 @@ maintainer_email  'help@sous-chefs.org'
 license           'Apache-2.0'
 description       'Installs and configures postgresql for clients or servers'
 long_description  IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version           '6.1.2'
+version           '7.0.0'
 source_url        'https://github.com/sous-chefs/postgresql'
 issues_url        'https://github.com/sous-chefs/postgresql/issues'
-chef_version      '>= 12.16' if respond_to?(:chef_version)
+chef_version      '>= 13.0'
 
 %w(ubuntu debian fedora amazon redhat centos scientific oracle).each do |os|
   supports os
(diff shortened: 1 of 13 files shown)


test/cookbooks/test/recipes/server_install.rb
```diff
@@ -8,7 +8,18 @@
   version '9.5'
 end
 
+pg_config = {
+  'default_statistics_target' => '100',
+  'maintenance_work_mem'      => "512MB",
```
martinisoft: Would switch the double quotes to single for consistency here

resources/server_install.rb
```diff
@@ -25,6 +25,8 @@
 property :port,              [String, Integer], default: 5432
 property :initdb_locale,     String, default: 'UTF-8'
 
+include PostgresqlCookbook::Helpers
```
martinisoft: Were these not included before? Why are they being added now?

metadata.rb
```diff
@@ -5,10 +5,10 @@
 license           'Apache-2.0'
 description       'Installs and configures postgresql for clients or servers'
 long_description  IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version           '6.1.2'
+version           '7.0.0'
```
martinisoft: Do we need to bump the version now?

damacus: Not really, but it reminds us/people that we're no longer aiming for 6.x series.

resources/server_install.rb
martinisoft: Ah. In that case I'd add it in to `action_class`.

damacus: The action_class block is the provider portion of the file. So it only works for properties methods in the `action` block. Now I think of it, that's probably why it's called the action class.

martinisoft: Oh right yeah you need that for a property, nevermind then.

PR #528, opened 2018-05-17
Damacus/patch libraries

damacus:
Reopening to fix server install issues


damacus:
Argh, because of the conflicts it might not be worth the merge.

PR #529, opened 2018-05-18
Damacus/tidy up tests

damacus:
### Description

Small tidy up of tests and suites

### Issues Resolved

- none

### Contribution Check List

- [ ] All tests pass.


.kitchen.dokken.yml
```diff
@@ -15,14 +15,8 @@ platforms:
   driver:
     image: dokken/amazonlinux
     pid_one_command: /sbin/init
-
-- name: debian-7
```
tas50: pour one out 🍾

resources/server_install.rb
```diff
@@ -30,6 +30,7 @@
 action :install do
   node.run_state['postgresql'] ||= {}
   node.run_state['postgresql']['version'] = new_resource.version
+  puts "HELLO DAN THE VERSION IS #{new_resource.version}"
```
tas50: ?

damacus:
erm debug?

resources/server_install.rb
damacus: I've removed it from latest, but I was debugging, some version numbers were... Wrong.

libraries/helpers.rb
```diff
@@ -1,7 +1,6 @@
 #
 # Cookbook:: postgresql
 # Library:: helpers
-# Author:: David Crane (<davidc@donorschoose.org>)
```
damacus: I've removed David as the Author here, as none of this code is actually his anymore. We've refactored so much of it. That the methods that he wrote no longer exist

PR #535, opened 2018-05-30
WIP - Issue 533

### LINKED ISSUE #533: postgresql_database idempotency
### Cookbook version
7.0.0

### Chef-client version
14.1.12

### Platform Details
Ubuntu 18.04

### Scenario:
Create a database

### Steps to Reproduce:
```ruby
postgresql_user 'test_user' do
  password  '123'
end

postgresql_database 'test_db' do
  owner 'test_user'
end
````

### Expected Result:
Database creation on first run, no action on subsequent runs.

### Actual Result:
It correctly creates the DB on the first run, but later fails to detect it:

```
STDERR: createdb: database creation failed: ERROR: database "database" already exists
```

It seems the same error as #514, creating a new one since version has changed and the other one is considered closed.

### PR DESCRIPTION
**Pull request by damacus**
### Description

Reworking library code 
Add idempotency to catch bad only_if/not_if code.
Add extension testing. So that we can test database and extension code

### Issues Resolved

#533 
#531 

### Contribution Check List

- [ ] All tests pass.

### DIFF
--- a/resources/access.rb
+++ b/resources/access.rb
@@ -16,14 +16,14 @@
 # limitations under the License.
 #
 
-property :source,        String, required: true, default: 'pg_hba.conf.erb'
-property :cookbook,      String, default: 'postgresql'
-property :comment,       [String, nil], default: nil
 property :access_type,   String, required: true, default: 'local'
 property :access_db,     String, required: true, default: 'all'
 property :access_user,   String, required: true, default: 'postgres'
-property :access_addr,   [String, nil], default: nil
 property :access_method, String, required: true, default: 'ident'
+property :cookbook,      String,                 default: 'postgresql'
+property :source,        String,                 default: 'pg_hba.conf.erb'
+property :access_addr,   String
+property :comment,       String
 
 action :grant do
   config_resource = new_resource
--- a/resources/extension.rb
+++ b/resources/extension.rb
@@ -16,14 +16,14 @@
 # limitations under the License.
 #
 
-property :database,    String, required: true
 property :extension,   String, name_property: true
+property :database,    String, required: true
 property :old_version, String
 
 action :create do
   create_query = "CREATE EXTENSION IF NOT EXISTS \"#{new_resource.extension}\""
   create_query << " FROM \"#{new_resource.old_version}\"" if property_is_set?(:old_version)
-  
+
   bash "CREATE EXTENSION #{new_resource.name}" do
     code psql_command_string(new_resource, create_query)
     user 'postgres'
--- a/resources/extension.rb
+++ b/resources/extension.rb
@@ -23,8 +23,9 @@ property :old_version, String
 action :create do
   create_query = "CREATE EXTENSION IF NOT EXISTS \"#{new_resource.extension}\""
   create_query << " FROM \"#{new_resource.old_version}\"" if property_is_set?(:old_version)
+  
   bash "CREATE EXTENSION #{new_resource.name}" do
-    code psql_command_string(new_resource.database, create_query)
+    code psql_command_string(new_resource, create_query)
     user 'postgres'
     action :run
     not_if { slave? || extension_installed? }
--- a/test/integration/access/controls/base_access.rb
+++ b/test/integration/access/controls/base_access.rb
@@ -12,6 +12,7 @@ control 'postgresl-local-access' do
   end
 
   postgres_access = postgres_session('postgres', '12345', '127.0.0.1')
+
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by damacus** [comment]
My test code is completely broken right now. But it's heading in the right direction.

**Comment by damacus** [comment]
Note to whomever wants to grab this next.

Right now it looks like users aren't being crated correctly.

**Comment by damacus** [comment]
@martinisoft Right now the `sous_chef` user can't login due to the password being wrong. 

I'm not sure what's going on here, but it's no longer able to login with the same access config. 🤔

**Comment by damacus** [comment]
```
# MAPNAME       SYSTEM-USERNAME         PG-USERNAME
testmap         postgres                postgres
```

```
# TYPE  DATABASE        USER            ADDRESS                 METHOD

# "local" is for Unix domain socket connections only
local   all             all                                     peer
# IPv4 local connections:
host    all             all             127.0.0.1/32            ident
# IPv6 local connections:
host    all             all             ::1/128                 ident
# Allow replication connections from localhost, by a user with the
# replication privilege.
#local   replication     postgres                                peer
#host    replication     postgres        127.0.0.1/32            ident
#host    replication     postgres        ::1/128                 ident
```

File: resources/server_install.rb [hunk 1]
```diff
@@ -44,6 +50,7 @@
     command rhel_init_db_command
     not_if { initialized? }
     only_if { platform_family?('rhel', 'fedora', 'amazon') }
+    notifies :write, 'log[Enable and start PostgreSQL service]', :immediately
```
damacus: This stops the log resource always firing.

damacus:
Im going to split out the extension fixes into another PR.

resources/access.rb
```diff
@@ -16,14 +16,14 @@
...
 
-property :source,        String, required: true, default: 'pg_hba.conf.erb'
-property :cookbook,      String, default: 'postgresql'
-property :comment,       [String, nil], default: nil
 property :access_type,   String, required: true, default: 'local'
 property :access_db,     String, required: true, default: 'all'
 property :access_user,   String, required: true, default: 'postgres'
-property :access_addr,   [String, nil], default: nil
 property :access_method, String, required: true, default: 'ident'
+property :cookbook,      String,                 default: 'postgresql'
```
sinfomicien: I read the source code... And i don't know where we use this parameter. Is it really needed to have and being able to modify it ?

resources/extension.rb
```diff
@@ -16,18 +15,41 @@
...
+property :user,     String, default: 'postgres'
+property :database, String, required: true
+property :host,     [String, nil]
+property :port,     Integer, default: 5432
 
 action :create do
-  create_query = "CREATE EXTENSION IF NOT EXISTS \"#{new_resource.extension}\""
-  create_query << " FROM \"#{new_resource.old_version}\"" if property_is_set?(:old_version)
+  extension_path = ::File.join(new_resource.source_directory, "#{new_resource.extension}#{new_resource.version}.sql")
+  cmd = %(psql -f "#{extension_path}" -d test_1 -U postgres --port 5432)
```
sinfomicien: Is it possible to use the full path of the psql command ? I know a bunch of admin who create system wide alias to force some options (who said RedHat maintainers ?)
I don't know if they are willing to do it on psql command... But ideally, i would try to use the real binay instead of a possible alias.

spec/libraries/helpers_spec.rb
```diff
@@ -78,4 +78,132 @@ class DummyClass < Chef::Node
...
+                             user: 'postgres',
+                             host: 'localhost',
+                             port: '5432'
+                            )
+      @query = 'THIS IS A COMMAND STRING'
+    end
+
+    it 'returns a full command string given all the parameters' do
+      grep_for = 'FOO'
+      result = %(psql -c "THIS IS A COMMAND STRING" -d db_foo -U postgres --host localhost --port 5432 | grep FOO)
```
sinfomicien: Same comment as before about the full path binary :-)

resources/access.rb
damacus: Ah right, everywhere we use a template, we should set  this. 

if it isn't used I should fix that! It allows the user to bring their own template.

sinfomicien: Misread the code, my bad, sorry

test/integration/access/controls/base_access.rb
```diff
@@ -1,8 +1,6 @@
 control 'postgresl-local-access' do
```
sinfomicien: Just because i wanna tell something. Better put postgres**q**l instead of postgresl  in the name of the control

PR #543, opened 2018-06-27
Updates to port usage.

jimdaga:
### Description
Updates to `port` usage.

**server_install**
- Removed duplicate `port` property.
- Changed `port` to only be an Integer.
- Fixed typo in word "preferences".

**server_conf**
- Added `port` as a property in **server_conf** as Integer and default 5432.
- Set `port` as a variable that gets passed to **postgresql.conf.erb**.
- Added `port` as something that gets added via **postgresql.conf.erb**.

_I found that changing `port` in server_install doesn't actually update the postgresql.conf file (it's mostly for the helper library when `psql_command_string` gets called). Using only **server_install** results in the default postgresql.conf file (aka 5432). You need to you need to use **server_conf** if you want another port. Users could set `port` in the `additional_config` hash, but to be consistant with all the other resources I think it would make sense that port is also a value that can be set in the **server config**._

**README**
- Removed the `port` property from **client_install** since it isn't valid value.
- Changed `port` to only be Integer for **server_install** and updated Example verbage.
- Added `port` to **server_conf** to make it clearer that this is how you actually change the listening port. 

### Issues Resolved

Didn't open one, let me know if I should.

### Contribution Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,6 @@ Name                | Types             | Description                           
 `ident_file`        | String            |                                                               | `#{conf_dir}/main/pg_ident.conf`          | no
 `external_pid_file` | String            |                                                               | `/var/run/postgresql/#{version}-main.pid` | no
 `password`          | String, nil       | Pass in a password, or have the cookbook generate one for you | 'generate'                                | no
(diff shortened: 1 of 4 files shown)


damacus:
This is quality. Nice one.

Issue #545, opened 2018-07-10
PostgreSQL is not the latest version


davidfetter:
### Cookbook version
All

### Chef-client version
All

### Platform Details
All

### Scenario:
Run the latest released version of PostgreSQL

### Steps to Reproduce:
1. Use the PostgreSQL cookbook
2. Log onto the node
3. psql -U postgres -AtXc 'SELECT version()'

### Expected Result:
PostgreSQL 10...

### Actual Result:
PostgreSQL 9.6...

### Proposed fix:
GH-544

damacus:
Looks like you haven't specified version 10 of postgresql, and just left the defaults?

PR #547, opened 2018-07-10
Cleanup metadata and improve client install test

martinisoft:
This is a change to make the client install test less brittle to version changes and improve the documentation of the test a little more.

I also noticed we aren't using the two cookbook dependencies in the metadata so I have removed those for even easier testing since Foodcritic will start screaming about build-essential in Chef 14.

While I was at it, I fixed the Amazon Linux builds and disabled the 18.04 Ubuntu builds because they simply were not completing. We might need @tas50's help on those.

* Helps with #545

- [x] All tests pass.

### DIFF
--- a/.kitchen.dokken.yml
+++ b/.kitchen.dokken.yml
@@ -14,8 +14,10 @@ platforms:
   driver:
     image: dokken/amazonlinux
     pid_one_command: /sbin/init
-    intermediate_instructions:
-      - RUN echo NETWORKING=yes > /etc/sysconfig/network
+    # intermediate_instructions:
+    #   - RUN echo NETWORKING=yes > /etc/sysconfig/network
+    # Note for future cokobook contributors: faking out the network like we do above breaks the network
+    # e.g. `getaddrinfo: Temporary failure in name resolution`
 
 - name: debian-8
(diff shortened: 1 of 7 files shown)


.kitchen.dokken.yml
```diff
@@ -14,6 +14,8 @@ platforms:
   driver:
     image: dokken/amazonlinux
     pid_one_command: /sbin/init
+    intermediate_instructions:
+      - RUN echo NETWORKING=yes > /etc/sysconfig/network
```
damacus: This breaks things. I've added and removed this line twice now. From here and Haproxy

README.md
```diff
@@ -32,9 +32,6 @@ The earliest supported version is currently:
 
 ### Cookbook Dependencies
 
-- `openssl`
-- `build-essential`
```
damacus: Good catch, I removed these ages ago

damacus:
aaand I changed my mind again (pre-coffee) 

This is why I stopped faking the network, whenever you fake out the network on amazon linux you get this failure whilst doing yum installs: 
`getaddrinfo: Temporary failure in name resolution`

PR #552, opened 2018-08-13
Fix HBA Order

skadz:
### Description

Fixes order of HBA file

### Issues Resolved

Allows users configuration to be read first, so that they are the rules used first (pg_hba.conf is read in order)

### Contribution Check List

- [X] All tests pass.
- [X] New functionality has been documented in the README if applicable

I don't think this needs updated documentation nor tests, as this just does the correct thing and has the defaults last instead of first (which causes the user to not be able to modify those settings).

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -5,7 +5,7 @@ maintainer_email  'help@sous-chefs.org'
 license           'Apache-2.0'
 description       'Installs and configures postgresql for clients or servers'
 long_description  IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version           '7.1.0'
+version           '7.1.1'
 source_url        'https://github.com/sous-chefs/postgresql'
 issues_url        'https://github.com/sous-chefs/postgresql/issues'
 chef_version      '>= 13.8'
--- a/templates/pg_hba.conf.erb
+++ b/templates/pg_hba.conf.erb
@@ -6,20 +6,10 @@
 # Refer to the "Client Authentication" section in the PostgreSQL
 # documentation for a complete description of this file.
 
-local   all             postgres                                peer
-
-# TYPE  DATABASE        USER            ADDRESS                 METHOD
-
-# "local" is for Unix domain socket connections only
-local   all             all                                     peer
-# IPv4 local connections:
-host    all             all             127.0.0.1/32            md5
-# IPv6 local connections:
-host    all             all             ::1/128                 md5
-
 ###########
 # From the postgresql_access resources
 ###########
+# TYPE  DATABASE        USER            ADDRESS                 METHOD
 <% @pg_hba.each do |k,v| -%>
 # <%= k %>
 <% if v[:comment] -%>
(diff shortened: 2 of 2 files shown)


metadata.rb
```diff
@@ -5,7 +5,7 @@
 license           'Apache-2.0'
 description       'Installs and configures postgresql for clients or servers'
 long_description  IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version           '7.1.0'
+version           '7.1.1'
```
martinisoft: Shouldn't have to bump the version here, please revert this.

skadz: Does that happen automatically with a release?  If not, I'm confused to why it wouldn't need a version bump.

martinisoft:
To elaborate further, `pg_hba` operates on a [first match basis](https://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html) so anything the user provides could override the defaults easily. I'd prefer this PR add an example to the README on how to use the cookbook and/or source properties of this resource to override the default order.

skadz:
@martinisoft  - As you just said, it's a first match, so when you have something like:

host    all             all             127.0.0.1/32            md5

And I need user foo to be able to connect without a password on localhost, I cannot do that, as it will first match this rule.

Another example, I need root to be able to access as the postgres user on local, with this rule:

local   all             postgres                                peer

I cannot do that, because this is the first match.

I understand these are good sane defaults, but as they are defaults, the user's configuration should come first and if they match no rules, then they hit these defaults.

damacus:
Closing the pr. But I'd suggest making another one with instructions how to feed in a new template to do exactly what you want ☺

martinisoft:
> I understand these are good sane defaults, but as they are defaults, the user's configuration should come first and if they match no rules, then they hit these defaults.

Hence why they are sane defaults that work for the 80% use case with users. If you've got a _really_ good reason to open your install up to the world then by all means, provide your own template, the functionality is there, but I'm not convinced this is something a user _expects_ to have out of the box.

skadz:
As a user, I fully expect that MY settings will override any defaults.  That's a pretty basic expectation.

martinisoft:
> As a user, I fully expect that MY settings will override any defaults. That's a pretty basic expectation.

Right but those defaults are generally from the packaging, which ones from the packaging don't you want? If that is the case, best you override the template. I'd want to see a failing test case here that proves the issue.

skadz:
Easy, allow root to access the postgres user locally.

psql -U postgres as root will never work with the current config

YpNo:
Hi,

I agree with @martinisoft. It is better to use your own template with "cookbook" parameter of the server_conf resource.
Otherwise, we can may be create an attribute or resource parameter to allow change authentication type (peer, ident, md5) of the local/host default entries. 

Regards.

metadata.rb
martinisoft: This is something we control as maintainers and Berkshelf is smart enough to source the current directory regardless of version for testing. The old even-odd release cycle in Chef hasn't been needed for a while.

damacus:
I'm going to lock this issue out now. But thank you for opening this issue,. 


If you feel this isn't completely resolved or you want to discuss it further please drop by the sous-chefs channel on the chef community slack.

PR #561, opened 2018-09-03
Document locale workaround for debian/ubuntu

### LINKED ISSUE #560: Locale needs defining [guidance]
## Cookbook version

7.1.0

## Chef-client version

14.4

## Platform Details

Any

## Scenario:

Create a database, specifying a different locale to en_US.UTF8

See here for an example fix:
https://github.com/sous-chefs/postgresql/blob/circleci/test/cookbooks/test/recipes/extension.rb#L5-L9

See #555 for more information

gionn:
fix #560

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,9 +105,8 @@ end
 
 #### Known issues
 
-On some platforms (e.g. Ubuntu 18.04), `initdb_locale` should be set to
-`en_US.utf8` instead of `en_US.UTF-8` due to
-[GH-555](https://github.com/sous-chefs/postgresql/issues/555).
+On some platforms (e.g. Ubuntu 18.04), your `initdb_locale` should be set to the
+same as the template database [GH-555](https://github.com/sous-chefs/postgresql/issues/555).
 
 ### postgresql_server_conf
 
@@ -323,9 +322,8 @@ end
 
 #### Known issues
 
-On some platforms (e.g. Ubuntu 18.04), `locale` should be set to `en_US.utf8`
-instead of `en_US.UTF-8` due to
-[GH-555](https://github.com/sous-chefs/postgresql/issues/555).
+On some platforms (e.g. Ubuntu 18.04), your `initdb_locale` should be set to the
+same as the template database [GH-555](https://github.com/sous-chefs/postgresql/issues/555).
 
 ### postgresql_user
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,8 +30,6 @@ The earliest supported version is currently:
 
 - Chef 13.8+
 
-### Cookbook Dependencies
```
damacus: Please remove this change, we want to be explicit that we don't require dependencies in this cookbook :)

README.md
```diff
@@ -101,6 +99,12 @@ postgresql_server_install 'Setup my postgresql 9.5 server' do
 end
 ```
 
+#### Known issues
+
+On Debian and Ubuntu platforms, `initdb_locale` should be set to `en_US.utf8`
```
damacus: I'm not sure this is debian/ubuntu specific. As in Docker/Dokken UTF-8 is the "standard" 

something with more generic might be better here :)

README.md
gionn: looked like missing documentation rather than no dependencies :+1:

README.md
```diff
@@ -18,7 +18,7 @@ If you are wondering where all the recipes went in v7.0+, or how on earth I use
 - Red Hat/CentOS/Scientific 6+
 - Fedora
 
-## PostgreSQL version
+### Postgresql version
```
damacus: When it isn't in all caps the PostgreSQL notes it with SQL capitalized
https://www.postgresql.org/about/

README.md
```diff
@@ -32,6 +32,8 @@ The earliest supported version is currently:
 
 ### Cookbook Dependencies
 
+None.
```
damacus: Maybe we should take this section out entirely?

README.md
```diff
@@ -313,6 +321,12 @@ postgresql_database 'my_app' do
 end
 ```
 
+#### Known issues
+
+On some platforms (e.g. Ubuntu 18.04), `locale` should be set to `en_US.utf8`
+instead of `en_US.UTF-8` due to
+[GH-555](https://github.com/sous-chefs/postgresql/issues/555).
```
damacus: I think the wording should be, something similar to 

> your initdb locale should be set to the same as the template database.

README.md
gionn: see https://github.com/sous-chefs/postgresql/pull/561#discussion_r214646862

PR #572, opened 2018-10-11
Removing notification unexistent property and replace examples

8la:
* notification property does not exist on the postgresql_* resources
* replace notification on the examples by common notifies...

### Description
Fix docs about notification betwen resources and postgresql service.

### Issues Resolved

-

### Contribution Check List

- [ ] All tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ To setup your PostgreSQL configuration with a specific data directory. If you ha
 postgresql_server_conf 'My PostgreSQL Config' do
   version '9.5'
   data_directory '/data/postgresql/9.5/main'
-  notification :reload
+  notifies :reload, 'service['postgresql']
 end
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,7 +137,7 @@ To setup your PostgreSQL configuration with a specific data directory. If you ha
 postgresql_server_conf 'My PostgreSQL Config' do
   version '9.5'
   data_directory '/data/postgresql/9.5/main'
-  notification :reload
+  notifies :reload, 'service['postgresql']
```
damacus: I actually, sadly, think this is still wrong. Thanks for the PR though. This service name doesn't always exist because distros like having their own names.

8la:
According to this: https://github.com/sous-chefs/postgresql/blob/master/resources/server_install.rb#L59

This should be right

PR #577, opened 2018-12-04
Fix setup postgresql extenstion

scopenco:
Hey,

current `postgresql_extension` skips installing extension because `shell_out` returns string "\n" for the absent extension instead of nil.

This PR fixes this issue.

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -71,7 +71,7 @@ module PostgresqlCookbook
       if new_resource.version
         version_result.stdout == new_resource.version
       else
-        !version_result.stdout.nil?
+        !version_result.stdout.eql?("\n")
       end
     end
 
--- a/test/cookbooks/test/recipes/extension.rb
+++ b/test/cookbooks/test/recipes/extension.rb
@@ -2,11 +2,11 @@ postgresql_repository 'install'
 
 postgresql_server_install 'package' do
   action [:install, :create]
-  initdb_locale 'en_US.utf8'
+  initdb_locale 'en_US.UTF-8'
 end
 
 postgresql_database 'test_1' do
-  locale 'en_US.utf8'
+  locale 'en_US.UTF-8'
 end
 
 package 'postgresql-contrib'
--- a/test/cookbooks/test/recipes/extension.rb
+++ b/test/cookbooks/test/recipes/extension.rb
@@ -1,3 +1,5 @@
+# recipes/extension.rb
+
 postgresql_repository 'install'
 
 postgresql_server_install 'package' do


test/cookbooks/test/recipes/extension.rb
```diff
@@ -2,11 +2,11 @@
 
 postgresql_server_install 'package' do
   action [:install, :create]
-  initdb_locale 'en_US.utf8'
+  initdb_locale 'en_US.UTF-8'
```
damacus: Is this actually what CircleCi/dokken expects? I put it lower case because no one can decide on which case to use... (and it *was* working as is)

libraries/helpers.rb
```diff
@@ -71,7 +71,7 @@ def extension_installed?(new_resource)
       if new_resource.version
         version_result.stdout == new_resource.version
       else
-        !version_result.stdout.nil?
+        !version_result.stdout.eql?("\n")
```
damacus: We should really be stripping `/n` rather than making sure something equals an invalid character

test/cookbooks/test/recipes/extension.rb
scopenco: I tried to converge it with testkitchen locally and it fails with locale 'en_US.utf8'.

damacus: It's probably a dokken vs kitchen thing and who built the images.

scopenco: OK, so removing?

damacus: can you push another commit please? I just poked CircleCI to run the tests on forked PRs :)

scopenco: pushed

kitchen-porter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add labels to this Pull Request</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a <a href="https://github.com/sous-chefs/line-cookbook/blob/master/CHANGELOG.md">CHANGELOG</a> entry.</td>
    </tr>
  </tbody>
</table>

</p>

damacus:
Ignore the labels, but could you add a changelog entry please?

damacus:
Looks like the extension fix failed in CI, if you don't get a chance to fix/debug today I'll take a look when I next can. 👍

aried3r:
It's weird that there's no test for this. Is there an easy to test way to make sure an extension was successfully installed in a database? Friendly ping @martinisoft

libraries/helpers.rb
martinisoft: I'd look at using `.chomp` to strip out extra whitespace which will give you a true on `empty?`

```ruby
"\n".chomp
=> ""
irb(main):002:0> "\n".chomp.empty?
=> true
```

aried3r:
Closing since this has been done in #595.

PR #582, opened 2018-12-17
Add Ruby source highlighting in README

aried3r:
Some more README corrections. Should have put them into a single PR :\

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,9 @@ Name                | Types             | Description                           
 
 #### Examples
 
-To install '9.5' version:
+To install version 9.5:
 
-```
+```ruby
 postgresql_client_install 'My PostgreSQL Client install' do
   version '9.5'
 end
@@ -89,15 +89,15 @@ Name                | Types           | Description                             
 
 #### Examples
 
-To install PostgreSQL server, set you own postgres password using non-default service port.
+To install PostgreSQL server, set your own postgres password using non-default service port.
 
-```
+```ruby
 postgresql_server_install 'My PostgreSQL Server install' do
   action :install
 end
 
-postgresql_server_install 'Setup my PostgreSQL 9.5 server' do
-  password 'MyP4ssw0d'
+postgresql_server_install 'Setup my PostgreSQL 9.6 server' do
+  password 'MyP4ssw0rd'
   port 5433
   action :create
 end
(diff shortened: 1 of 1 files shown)


kitchen-porter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please provide a summary of your Pull Request.</td>
    </tr>
  </tbody>
</table>

</p>

README.md
```diff
@@ -89,15 +89,15 @@ Name                | Types           | Description
...
-```
+```ruby
 postgresql_server_install 'My PostgreSQL Server install' do
   action :install
 end
 
-postgresql_server_install 'Setup my PostgreSQL 9.5 server' do
-  password 'MyP4ssw0d'
+postgresql_server_install 'Setup my PostgreSQL 9.6 server' do
+  password 'MyP4ssw0rd'
```
aried3r: Just couldn't look at this without the "r" :D

README.md
```diff
@@ -89,15 +89,15 @@ Name                | Types           | Description
...
 
-```
+```ruby
 postgresql_server_install 'My PostgreSQL Server install' do
   action :install
 end
 
-postgresql_server_install 'Setup my PostgreSQL 9.5 server' do
-  password 'MyP4ssw0d'
+postgresql_server_install 'Setup my PostgreSQL 9.6 server' do
```
aried3r: 9.6 is currently the default, so that's the correct name I think.

JohnRoesler: do we want to update all the examples to 9.6 then?

aried3r: Well, the _default_ is 9.6 and this example doesn't define a version, that's why I wrote 9.6 here. TBH it might be best to remove the version here because it's just a string and chances that this won't get maintained when the default changes are probably very high.

Where 9.5 is used it's to show that different versions can be installed as well. Personally I'd change those to 10.6 or 11.1
https://www.postgresql.org/support/versioning/

On that note, the README says we support 9.3, while 9.3 is now officially unsupported. I'm not sure what to do in this case. Either the README should not mention specific version numbers (since I saw no tests actually testing 9.3 or even 10 or 11) or we need to release an update?

damacus: Yeah we totally don't support 9.3 it should just say we support https://www.postgresql.org/support/versioning/

which is what I thought I put! :(

damacus: Ah it does. But I then say, we support this minor version. 

Can you drop that bit off to please?

PR #593, opened 2019-01-14
Update documentation about generating passwords

aried3r:
Before it looked like it might be the hardcoded string "generate".

WDYT?


damacus:
This is much clearer thanks.

I think it used to generate one for you,hence why this was there

PR #630, opened 2019-10-24
Feature/add encoding initdb

### LINKED ISSUE #596: Cannot set initdb_locale on Ubuntu/Debian
### Cookbook version
7.1.2

### Chef-client version
14.8.12

### Platform Details
Ubuntu 16.04 LTS

### Scenario:
Trying to set `initdb_locale` for repo-installation of PostgreSQL.

### Steps to Reproduce:
```Ruby
postgresql_server_install 'Install server from Repo' do
  action     :install
  version    POSTGRESQL_VERSION
  password   'MyP4$$w0rd'
  setup_repo true
  initdb_locale 'de_DE.utf-8'
end
```

### Expected Result:
The `template1` DB that is being initialized by `initdb` should have a `de_DE.utf-8` locale but is using the system locale instead.

The `initdb_locale` value is only used for some linux distros as well:

https://github.com/sous-chefs/postgresql/blob/a66d05c02b2e3e3a3841ea6dcdb5be55c0e8d7da/resources/server_install.rb#L48-L53
https://github.com/sous-chefs/postgresql/blob/a66d05c02b2e3e3a3841ea6dcdb5be55c0e8d7da/libraries/helpers.rb#L206-L217

### Actual Result:
On Debian/Ubuntu using the repo versions the initdb locale is set automatically to the system locale (if I understood the links below correctly), in this case `en_US.utf8` because that's what we use on our host system. We could try to overwrite the `LC_*` environment variables to manipulate this, but instead what we're currently doing at work is using an explicit locale when creating databases, using the `template0` database instead of `template1`.

> For a given locale category, say the collation, the following environment variables are consulted in this order until one is found to be set: `LC_ALL`, `LC_COLLATE` (or the variable corresponding to the respective category), `LANG`. <sup>[source](https://www.postgresql.org/docs/10/locale.html)</sup>

> Debian PostgreSQL installation automatically calls the initdb i.e. it initializes the cluster with default encoding and locale. Encoding can be changed later but the locale cannot. <sup>[source](https://wiki.debian.org/PostgreSql)</sup>

The above quote from the debian wiki might not exactly describe what's happening here, since `setup_repo` used the PGPD and not the official debian packages, but it's what I am experiencing.

JanStevens:
# Description

Ubuntu provides postgresql-common to manage multiple pg clusters on one host. This comes with a createcluster.conf file that determines how a new cluster is created. By default when installing the postgres server package a new cluster is created using the createcluster.conf file

## Issues Resolved

#596 

## Check List


damacus:
@JanStevens thanks for the contribution. I've been away for a while, but I'll take a look at getting this merged in this week!

damacus:
Looks like this is genuinely broken. 

Running dokken test (kitchen with the local Dokken driver) all the extension test fail locally. 

It might be something to do with dokken networking, but I'm going to leave that as an excersice for the reader 👍

JanStevens:
> Looks like this is genuinely broken.
> 
> Running dokken test (kitchen with the local Dokken driver) all the extension test fail locally.
> 
> It might be something to do with dokken networking, but I'm going to leave that as an excersice for the reader 👍

Yea, Ill see whats up, I do remember that I got stuck and hoped another sous chef could help out

damacus:
I'm going to do some work on the release side of things first, and then swing back around to this. 

It might be worth prying apart dokken from this change and getting it working in normal, Vagrant/Test Kitchen first. Though that's going to take a fix of this encoding section, you'll probably need to detect which hypervisor you're using and use the correct locale for it, or something similar 

https://github.com/sous-chefs/postgresql/blob/master/test/cookbooks/test/recipes/extension.rb#L4-L8

```ruby
encoding = value_for_platform(
  [:debian, :ubuntu, :fedora] => { default: 'C.UTF-8' },
  centos: { default: 'en_GB.utf8' },
  default: 'en_US.utf8'
)
```

Without a fix for that, the extension test only works for Dokken, which is our main platform for tesing.

JanStevens:
All green, it seemed that I mixed up encoding vs locale in that spec

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This file is used to list changes made in the last 3 major versions of the postg
 ## v7.1.5 (2019-11-18)
 
 - Allow to install extensions with hyphens, ex: `postgresql_extension '"uuid-ossp"'`
+- Correctly configure postgres-common on ubuntu hosts, fixes #596
```
Xorima: Can you pop this under the `## unreleased` grouping at the top (or make one if it doesn't exist)

Issue #653, opened 2020-06-05
update_user_with_attributes_sql is broken


Zhann:
### :speaking_head: Foreword

Thank for taking the time to fill this bug report fully. Without it we may not be able to fix the bug, and the issue may be closed without resolution.

## :ghost: Brief Description

[update_user_with_attributes_sql](https://github.com/sous-chefs/postgresql/blob/v7.1.9/libraries/helpers.rb#L130-L133) cannot work, `attr` is never passed to the method, causing a NameError:

```
undefined local variable or method `attr' for #<#<Class:0x000000000432a3f8>:0x0000000003ea73a8>
```

## :pancakes: Cookbook version

v7.1.9

## :woman_cook:  Chef-Infra Version

using 14.15.6-1, but I assume all

## :tophat: Platform details

all

## Steps To Reproduce

Run the following code:

```
postgresql_user 'foo' do
  action %i[
    create
    update
  ]

  password 'bar'
  attributes({ statement_timeout: '8mins' })
end

```

## :police_car: Expected behavior

Update user with attributes

PR #654, opened 2020-06-05
Passes attr to update_user_with_attributes_sql

Zhann:
# Description

Passes `attr` to `update_user_with_attributes_sql` to make updates on `postgresql_user` work with attribute hashes.

Also removes quotes from rolname, as that's a syntax error in postgresql.

## Issues Resolved

Fixes https://github.com/sous-chefs/postgresql/issues/653

## Check List

- [ ] ~~New functionality has been documented in the README if applicable~~.

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -128,7 +128,7 @@ module PostgresqlCookbook
     end
 
     def update_user_with_attributes_sql(new_resource, attr, value)
-      sql = %(ALTER ROLE '#{new_resource.create_user}' SET #{attr} = #{value})
+      sql = %(ALTER ROLE #{new_resource.create_user} SET #{attr} = #{value})
       psql_command_string(new_resource, sql)
     end
 
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -127,7 +127,7 @@ module PostgresqlCookbook
       psql_command_string(new_resource, sql)
     end
 
-    def update_user_with_attributes_sql(new_resource, value)
+    def update_user_with_attributes_sql(new_resource, attr, value)
       sql = %(ALTER ROLE '#{new_resource.create_user}' SET #{attr} = #{value})
       psql_command_string(new_resource, sql)
     end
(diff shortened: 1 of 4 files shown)


damacus:
Hey there!

Thanks for the PR. It might be worth putting a test in to exercise this. 

Cheers,
Dan

Zhann:
Thanks @damacus, added an integration test. Moved it to the access integration tests, not sure if it makes sense to put it there, but it felt like the closest fit. Let me know if you think I should do it differently.

libraries/helpers.rb
```diff
@@ -127,8 +127,8 @@ def update_user_sql(new_resource)
       psql_command_string(new_resource, sql)
     end
 
-    def update_user_with_attributes_sql(new_resource, value)
-      sql = %(ALTER ROLE '#{new_resource.create_user}' SET #{attr} = #{value})
+    def update_user_with_attributes_sql(new_resource, attr, value)
+      sql = %(ALTER ROLE #{new_resource.create_user} SET #{attr} = #{value})
```
Xorima, line 131: Question about this change, (not a postgresql user) is it possible to create a user with a space in the name?

Xorima:
Hey @Zhann 

Could you take a look at the failing test, should be a simple fix:

```
  ×  sous_chef has statement_timeout set to 8mins: PostgreSQL query with errors: SHOW statement_timeout;
     ×  PostgreSQL query with errors: SHOW statement_timeout; output is expected to cmp == /8min/
     
     expected: (?-mix:8min)
          got: 
     psql: FATAL:  database "sous_chef" does not exist
     
     
     (compared using `cmp` matcher)

```

Thanks

libraries/helpers.rb
Zhann: I don't think you can (or at least, it's not the smartest idea). You can quote names btw, but then you have to use double quotes, single quotes a syntax error.

I defaulted to the style of other methods in the helper file, assuming that made most sense.

Zhann:
@Xorima fixed, thank you!

Zhann:
16.04 machines seems to fail on `server_install.rb`, but that's not really in scope of (or related to) this change.

Zhann:
Thank you for merging 🙇 

will you be creating a new release for this? (7.1.10)

Xorima:
I will leave for @damacus so he can test the auto release system

Issue #629, opened 2019-10-15
Error on postgresql_user action:drop


PhilippeBerard:
### :speaking_head: Foreword

To deploy our customers' SaaS applications, we use this cookbook for creating the corresponding database and database user.

## :ghost: Brief Description

After droping a database, droping the user systematically fails with a "ShellCommandFailed: Command execution failed". Droping the user manually works flawlessly.
 
## :pancakes: Cookbook version

7.1.4

## :woman_cook:  Chef-Infra Version

14.11.21

## :tophat: Platform details

Debian 9 Stretch
Kernel 4.9.0-9-amd64
Linux jalios29 4.9.0-9-amd64 #1 SMP Debian 4.9.168-1+deb9u4 (2019-07-19) x86_64 GNU/Linux

## Steps To Reproduce

1/ Create User
2/ Create Database
3/ Delete Database
4/ Delete User
5/ ERROR

## :police_car: Expected behavior

Database user dropped without error.

## :heavy_plus_sign: Additional context

The machines where the error occur handle Apache2 websites, Tomcat 8 webapps with OpenJDK 8.

PhilippeBerard:
Would it be possible to have any return on this issue ? Thanks.

damacus:
Hey @PhilippeBerard sorry about the the lack of response on this issue, and thanks for sending a reminder. 

Are you able to write a test in a PR to show the error? That way we have an test that we can confirm works to stop future regression. Thanks!

JonBetts:
I'm getting an issue dropping users on version 7.1.9 of the cookbook. Looks like it is putting single quotes around the user/role name in the statement:

```
---- Begin output of /usr/bin/psql -c "DROP ROLE IF EXISTS 'jon'" -U postgres --port 5432 ----
STDOUT: 
STDERR: ERROR:  syntax error at or near "'jon'"
LINE 1: DROP ROLE IF EXISTS 'jon'
                            ^
---- End output of /usr/bin/psql -c "DROP ROLE IF EXISTS 'jon'" -U postgres --port 5432 ----
```

Not sure if this needs a test or anything, as it sounds like a potentially simple fix.

Zhann:
> I'm getting an issue dropping users on version 7.1.9 of the cookbook. Looks like it is putting single quotes around the user/role name in the statement:
> 
> ```
> ---- Begin output of /usr/bin/psql -c "DROP ROLE IF EXISTS 'jon'" -U postgres --port 5432 ----
[...]

PR #661, opened 2020-09-08
Removes quotes around user in DROP ROLE IF EXISTS

Zhann:
Fixes https://github.com/sous-chefs/postgresql/issues/629

# Description

Fixes sql syntax by removing quotes around username in DROP ROLE IF EXISTS

## Issues Resolved

Resolves https://github.com/sous-chefs/postgresql/issues/629

## Check List

- [ ] ~~New functionality has been documented in the README if applicable.~~

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -139,7 +139,7 @@ module PostgresqlCookbook
     end
 
     def drop_user_sql(new_resource)
-      sql = %(DROP ROLE IF EXISTS '#{new_resource.create_user}')
+      sql = %(DROP ROLE IF EXISTS #{new_resource.create_user})
       psql_command_string(new_resource, sql)
     end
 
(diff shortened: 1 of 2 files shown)


ramereth:
@Zhann do you think you could add something to the integration recipe/tests to ensure this is actually fixed? That would be ideal to ensure this works properly and doesn't happen in the future.

Zhann:
> @Zhann do you think you could add something to the integration recipe/tests to ensure this is actually fixed? That would be ideal to ensure this works properly and doesn't happen in the future.

I think so, yes. added a user with actions `action [:create, :drop]`.

ramereth:
@Zhann looks like it actually produces an error which is good we caught it. Can you please resolve it?

ramereth:
Closing in favor of #672

Issue #676, opened 2021-01-26
The psqlrc files should not be read by helper functions


mimor:
## :ghost: Brief Description

The [psql_command_string function](https://github.com/sous-chefs/postgresql/blob/9094a200ce0fa1289580e9f546f8854066ec5580/libraries/helpers.rb#L24) will read the [psqlrc files](https://wiki.postgresql.org/wiki/Psqlrc) on a system as per the default.

When the psqlrc files contain entries that produce output, such as timestamps, they can break functionality when being parsed by a conditional such as `not_if`.

## :pancakes: Cookbook version

8.1.1

## :woman_cook:  Chef-Infra Version

14.14.29

## :tophat: Platform details

Ubuntu 18.04

## Steps To Reproduce

Steps to reproduce the behaviour:

1. Set a /etc/postgresql-common/psqlrc on your system with the following content:
```
\timing
\set PROMPT1 'foo# '
```
2. Use a function that uses the psql_command_string in conjunction with not_if, for example:
```
postgresql_database 'foo' do
  owner 'foo'
end
```
3. This will trigger `not_if { database_exists?(new_resource) }` and thus skip the creation, even if it does not exit.

## :police_car: Expected behavior

The database foo should be created.
The psqlrc file should be ignored by the helpers.

## :heavy_plus_sign: Additional context

A PR is in the making.

PR #677, opened 2021-01-26
[GH-676] Disables psqlrc for helper functions

mimor:
The psqlrc files should not be read by helperscripts.

Entries such as `/timing` will introduce unexpected output, potentially failing
`not_if` functionality when invoking these helpers.

Ref: https://github.com/sous-chefs/postgresql/issues/676

Signed-off-by: Mike Morraye <hello@morraye.be>

# Description

Describe what this change achieves

## Issues Resolved

## Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This file is used to list changes made in the last 3 major versions of the postg
 
 ## Unreleased
 
+## Unreleased
+
+- Makes psqlrc optional when invoking `psql_command_string`
+
 ## 8.1.1 - *2021-01-12*
 
 - Fix attribute updates for users with dashes
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -28,6 +28,7 @@ module PostgresqlCookbook
       cmd << " --host #{new_resource.host}" if new_resource.host
       cmd << " --port #{new_resource.port}" if new_resource.port
       cmd << ' --tuples-only'               if value_only
+      cmd << ' --no-psqlrc'                 unless new_resource.psqlrc
       cmd << " | grep #{grep_for}"          if grep_for
       cmd
     end
--- a/resources/database.rb
+++ b/resources/database.rb
@@ -26,6 +26,7 @@ property :user,     String, default: 'postgres'
 property :database, String, name_property: true
 property :host,     String
 property :port,     Integer, default: 5432
+property :psqlrc,   [true, false], default: true
 
 action :create do
   createdb = 'createdb'
--- a/resources/extension.rb
+++ b/resources/extension.rb
@@ -25,6 +25,7 @@ property :user,     String, default: 'postgres'
 property :database, String, required: true
 property :host,     [String, nil]
 property :port,     Integer, default: 5432
+property :psqlrc,   [true, false], default: true
 
 action :create do
   bash "CREATE EXTENSION #{new_resource.name}" do
--- a/resources/server_install.rb
+++ b/resources/server_install.rb
@@ -32,6 +32,7 @@ property :initdb_encoding,   String
 property :user,     String, default: 'postgres'
 property :database, String
 property :host,     [String, nil]
+property :psqlrc,   [true, false], default: true
 
 action :install do
   node.run_state['postgresql'] ||= {}
--- a/resources/user.rb
+++ b/resources/user.rb
@@ -33,6 +33,7 @@ property :user,     String, default: 'postgres'
 property :database, String
 property :host,     String
 property :port,     Integer, default: 5432
+property :psqlrc,   [true, false], default: true
 
 action :create do
   Chef::Log.warn('You cannot use "attributes" property with create action.') unless new_resource.attributes.empty?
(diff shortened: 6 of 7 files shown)


kitchen-porter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a CHANGELOG entry.</td>
    </tr>
  </tbody>
</table>

</p>

libraries/helpers.rb
```diff
@@ -28,6 +28,7 @@ def psql_command_string(new_resource, query, grep_for: nil, value_only: false)
       cmd << " --host #{new_resource.host}" if new_resource.host
       cmd << " --port #{new_resource.port}" if new_resource.port
       cmd << ' --tuples-only'               if value_only
+      cmd << ' --no-psqlrc'
       cmd << " | grep #{grep_for}"          if grep_for
```
ramereth, line 32: I think this might resolve the cookstyle issue.
```suggestion
      cmd << " --no-psqlrc | grep #{grep_for}" if grep_for
```

ramereth (changes requested):
Please also add a Changelog entry.

libraries/helpers.rb
mimor: Thank you for the review.

I had the impression from the [CONTRIBUTING.md](https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD#code-review-process) that the changelog would be added at a later stage, during the review process 😅
My impression was wrong.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This file is used to list changes made in the last 3 major versions of the postg
 
 ## Unreleased
 
+## 8.1.2 - 2021-01-27
```
ramereth, line 7: Please use this header as our automated release system will fix this when it's released.
```suggestion
## Unreleased
```

libraries/helpers.rb
ramereth: Ah  yes, we probably need to amend that.

libraries/helpers.rb
```diff
@@ -28,7 +28,8 @@ def psql_command_string(new_resource, query, grep_for: nil, value_only: false)
       cmd << " --host #{new_resource.host}" if new_resource.host
       cmd << " --port #{new_resource.port}" if new_resource.port
       cmd << ' --tuples-only'               if value_only
-      cmd << " | grep #{grep_for}"          if grep_for
+      cmd << ' --no-psqlrc'
```
ramereth, line 31: Thinking about this more, I think it might be a good idea to make this a new parameter called `psqlrc` which is enabled by default (which mimics the current behavior), but folks like you can disable it.

I can just see this impacting folks who are expecting this to work like it has been.

mimor: There was an attempt to get a boolean value from node attributes such as `node['postgresql']['no_psqlrc']` as to have a way to change the behaviour of the already defined helpers, but I did not succeed... 😞

ramereth: This would be in the resource and not done via attributes. We're trying to move away from attributes towards more resource driven cookbooks so we don't run into issues.  Would you like some help or give an example of how to do this?

mimor: That would be great! 🙇

ramereth: Here's a diff with an example. You'd want to add this property to every resource that also has a `host` property.
```diff
diff --git a/libraries/helpers.rb b/libraries/helpers.rb
index 911a094..d473ed0 100644
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -21,14 +21,14 @@ module PostgresqlCookbook
 
     require 'securerandom'
 
-    def psql_command_string(new_resource, query, grep_for: nil, value_only: false, no_psqlrc: false)
+    def psql_command_string(new_resource, query, grep_for: nil, value_only: false)
       cmd = %(/usr/bin/psql -c "#{query}")
       cmd << " -d #{new_resource.database}" if new_resource.database
       cmd << " -U #{new_resource.user}"     if new_resource.user
       cmd << " --host #{new_resource.host}" if new_resource.host
       cmd << " --port #{new_resource.port}" if new_resource.port
       cmd << ' --tuples-only'               if value_only
-      cmd << ' --no-psqlrc'                 if no_psqlrc
+      cmd << ' --no-psqlrc'                 unless new_resource.psqlrc
       cmd << " | grep #{grep_for}"          if grep_for
       cmd
     end
diff --git a/resources/database.rb b/resources/database.rb
index c5307fb..a17d5c3 100644
--- a/resources/database.rb
+++ b/resources/database.rb
@@ -26,6 +26,7 @@ property :user,     String, default: 'postgres'
 property :database, String, name_property: true
 property :host,     String
 property :port,     Integer, default: 5432
+property :psqlrc,   [true,false], default: true
 
 action :create do
   createdb = 'createdb'
```

mimor: 🙇 🙇 🙇 I have used your example and fixed my PR accordingly.

ramereth (approved):

kitchen-porter:
Released as: [8.2.0](https://github.com/sous-chefs/postgresql/releases/tag/8.2.0)

PR #682, opened 2021-03-23
Fix idempotency when installing multiple client versions

ramereth:
This fixes an issue when trying to install multiple versions of postgresql on the same system. This is useful on backup servers where you might be backing up several different versions and it's best to use the correct version when doing that.

On RHEL-based systems, this changes the name of the gpg key to *not* include the PostgreSQL version. Each repo uses the same key so this shouldn't be a problem.

On Debian-based systems, create a different repo based on the version you're using so that it also works on those systems.

Signed-off-by: Lance Albertson <lance@osuosl.org>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,7 @@ jobs:
           - 'client-install-12'
           - 'client-install-95'
           - 'client-install-96'
+          - 'client-multi-install'
           - 'extension-10'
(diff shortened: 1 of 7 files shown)


damacus (approved):

kitchen-porter:
Released as: [8.3.0](https://github.com/sous-chefs/postgresql/releases/tag/8.3.0)

Issue #702, opened 2022-03-21
Readme needs required Chef version bumping to 16.0+


KillerCars:
### :speaking_head: Foreword

Thank for taking the time to fill this bug report fully. Without it we may not be able to fix the bug, and the issue may be closed without resolution.

## :ghost: Brief Description

README.md shows Chef 15.3+, but should be 16.0+

## :pancakes: Cookbook version

Version of the cookbook where you are encountering the issue.

v10.0.0+

## :woman_cook:  Chef-Infra Version

Version of chef-client in your environment.

15.14

## :tophat: Platform details

Operating system distribution and release version. Cloud provider if running in the cloud.

## Steps To Reproduce

1. Go to 'README.md'
2. Scroll down to 'https://github.com/sous-chefs/postgresql#chef'
3. Look with your eyes

## :police_car: Expected behavior

Should read '16.0+'

## :heavy_plus_sign: Additional context

Add any other context about the problem here. e.g. related issues or existing pull requests.

KillerCars:
I tried to raise a PR for this but got a 403.

p.s. I am a total github noob

Issue #701, opened 2022-02-03
Permit yum options to be passed via recipes


ekelson-bcove:
### :speaking_head: Foreword
It would be wonderful if we could pass yum options when install features.  If you have say a repo disabled by default, there is no method for passing the --enablerepo option.

## :person_frowning: Problem Statement

Without this feature, this cookbook is unusable to those whose repos are disabled by default

## :grey_question: Possible Solution

Copy the pattern used by the package function in chef

## :arrow_heading_up: Describe alternatives you've considered

We could stop disabling repos, however it would be great for this cookbook to follow conventions set by other chef cookbooks

## :heavy_plus_sign: Additional context

none

Issue #695, opened 2021-09-07
postgresql_extension install fails - NoMethodError: undefined method `[]' for nil:NilClass


noellowry:
## :ghost: Brief Description
Trying to install extension but fails with NoMethodError: undefined method `[]' for nil:NilClass

## :pancakes: Cookbook version

9.0.3

## :woman_cook:  Chef-Infra Version

17.4.38

## :tophat: Platform details

vSphere

## :police_car: Expected behavior
Expect that extension gets installed but it ends in error

```
postgresql_extension 'pg_stat_statements' do
    database 'postgres'
end

```
```
17:43:52  [0;32m    vsphere-iso:   * postgresql_extension[pg_stat_statements] action create[0m
17:43:52  [0;32m    vsphere-iso:     * bash[CREATE EXTENSION pg_stat_statements] action run[0m
17:43:52  [0;32m    vsphere-iso:[0m
17:43:52  [0;32m    vsphere-iso:       ================================================================================[0m
17:43:52  [0;32m    vsphere-iso:       Error executing action `run` on resource 'bash[CREATE EXTENSION pg_stat_statements]'[0m
17:43:52  [0;32m    vsphere-iso:       ================================================================================[0m
17:43:52  [0;32m    vsphere-iso:[0m
17:43:52  [0;32m    vsphere-iso:       NoMethodError[0m
17:43:52  [0;32m    vsphere-iso:       -------------[0m
17:43:52  [0;32m    vsphere-iso:       undefined method `[]' for nil:NilClass[0m
17:43:52  [0;32m    vsphere-iso:[0m
17:43:52  [0;32m    vsphere-iso:       Cookbook Trace: (most recent call first)[0m
17:43:52  [0;32m    vsphere-iso:       ----------------------------------------[0m
17:43:52  [0;32m    vsphere-iso:       /tmp/packer-chef-solo/local-mode-cache/cache/cookbooks/postgresql/libraries/helpers.rb:147:in `data_dir'[0m
17:43:52  [0;32m    vsphere-iso:       /tmp/packer-chef-solo/local-mode-cache/cache/cookbooks/postgresql/libraries/helpers.rb:175:in `follower?'[0m
17:43:52  [0;32m    vsphere-iso:       /tmp/packer-chef-solo/local-mode-cache/cache/cookbooks/postgresql/resources/extension.rb:36:in `block (3 levels) in class_from_file'[0m
[...]

Issue #694, opened 2021-09-02
external_pid_file and stats_temp_directory set incorrectly when installing different version than 12 (default)


hrak:
## :ghost: Brief Description

When installing a different version than is the default, `external_pid_file` and `stats_temp_directory` are set incorrectly by default. The only way to fix this is to set them explicitly.

## :pancakes: Cookbook version

9.0.3

## :woman_cook:  Chef-Infra Version

17.4.18

## :tophat: Platform details

Ubuntu 20.04

## Steps To Reproduce

1. `kitchen converge server-install-96-ubuntu-2004`
2. `kitchen login`
3. `cat /etc/postgresql/9.6/main/postgresql.conf`
4. See version 12 mentioned in external_pid_file and stats_temp_directory

```
data_directory = '/var/lib/postgresql/9.6/main'
hba_file = '/etc/postgresql/9.6/main/pg_hba.conf'
ident_file = '/etc/postgresql/9.6/main/pg_ident.conf'
external_pid_file = '/var/run/postgresql/12-main.pid'
stats_temp_directory = '/var/run/postgresql/12-main.pg_stat_tmp'
port = 5432
```

## :police_car: Expected behavior

`external_pid_file` & `stats_temp_directory` pointing to version 9.6

PR #705, opened 2022-12-04
v11 Table Flip

bmhughes:
# Description

- Major refactor, see [UPGRADING.md](UPGRADING.md)
- Condense repository, client_install and server_install into common install resource with actions replacing the previous split resources.
  - `:repository` action
  - `:install` action
  - `:client_install` action
  - `:server_install` action
- Refactor database, role (user) and extension resources to use pg gem for database access
  - All `load_current_value` support
- Rename user resource to role to match PostgreSQL (still aliased to `:postgresql_user`)
- Update various resource properties to match PostgreSQL side

## Issues Resolved

- #702 
- #701 
- #695 
- #674 
- #694
- #648 
- #668 
- #642 
- #605 
- #602 
- #573

## Check List


libraries/access.rb
```diff
@@ -0,0 +1,293 @@
...
+
+        class PgHbaFile
+          attr_reader :entries
+
+          def initialize
+            @entries = []
+            @access_entries = []
+          end
+
+          def read!(file = 'pg_hba.conf', sort: true)
```
damacus, line 74: Is read! Going to modify the file?

bmhughes: Other way around it'll load from the file and clobber whatever was in the object before

Xorima:
Can you draft this @bmhughes until it is ready ?

bmhughes:
> Can you draft this @bmhughes until it is ready ?

Can we run actions on a draft?

Stromweld:
yes they still run, but the emails and notifications are suppressed on updatest to the PR till taken out of draft status.

Stromweld:
It'll also block merging until taken out of draft status as well.

bmhughes:
Hmm I didn't think they were running before, let me try.

spec/spec_helper.rb
```diff
@@ -1,4 +1,3 @@
-# frozen_string_literal: true
 require 'chefspec'
 require 'chefspec/berkshelf'
```
Stromweld, line 2: Since you have this major update should we switch this cookbook to policyfile vs berkshelf? I think that'd polish this off nicely and make it a great example for community on how to write a modern cookbook, resources, libraries, etc...

Stromweld (changes requested):
Looks purty

spec/spec_helper.rb
bmhughes: Personally, I'm not a massive fan of policyfile for library cookbooks (or policyfile full stop as an aside). I'm not sure what it'd give the cookbook that it doesn't get from berks?

Stromweld: Other than future proofing it really doesn’t give anything over berks especially for library cookbooks.

Stromweld (approved):
Awesome job

libraries/access.rb
damacus: If it isn't going to modify it then it doesn't need the !

spec/spec_helper.rb
damacus: I think until it breaks we should leave all the cookbooks on Berks. See my haproxy PR. There was little benefit. 
But if we do it should be it's own pr

Stromweld: I'm fine with that, just brought it up more for discussion.

libraries/access.rb
bmhughes: Hmm i'm not so sure as it modifies the receiver by replacing any entries in the object with the ones in the file, removing any changes. The same as `:map!` will change whatever it's called upon.

kitchen-porter:
Released as: [11.0.0](https://github.com/sous-chefs/postgresql/releases/tag/11.0.0)

PR #710, opened 2022-12-15
Allow installing native distribution packages

### LINKED ISSUE #709: Unable to use the distribution packages
## :ghost: Brief Description

Before version 11 of the cookbook, there was a `setup_repo` property to choose whether to use upstream or distribution packages of PostgreSQL. This property was transformed to the `source` property. The new property doesn't allow to use distribution packages anymore.

## :pancakes: Cookbook version

11.0.1

## :woman_cook:  Chef-Infra Version

17.10.0

## :tophat: Platform details

Debian 11

## Steps To Reproduce

1. Consider the following recipe

```ruby
postgresql_install 'postgresql' do
  version 13
  source :nil
  action :install
end
```

2. Converge

```
       Cinc Client, version 17.10.0
       Patents: https://www.chef.io/patents
       Infra Phase starting
       Creating a new client identity for test-debian-11 using the validator key.
       Resolving cookbooks for run list: ["apt::default", "test::test"]
       Synchronizing cookbooks:
         - apt (7.4.3)
         - postgresql (11.0.1)
         - test (1.0.0)
         - yum (7.4.1)
       Installing cookbook gem dependencies:
       Compiling cookbooks...
       Loading Cinc Auditor profile files:
       Loading Cinc Auditor input files:
       Loading Cinc Auditor waiver files:
       
       ================================================================================
       Recipe Compile Error in /tmp/kitchen/cache/cookbooks/test/recipes/test.rb
       ================================================================================
       
       Chef::Exceptions::ValidationFailed
       ----------------------------------
       Option source must be equal to one of: :repo!  You passed :nil.
       
       Cookbook Trace: (most recent call first)
       ----------------------------------------
         /tmp/kitchen/cache/cookbooks/test/recipes/test.rb:3:in `block in from_file'
         /tmp/kitchen/cache/cookbooks/test/recipes/test.rb:1:in `from_file'
       
       Relevant File Content:
       ----------------------
       /tmp/kitchen/cache/cookbooks/test/recipes/test.rb:
       
         1:  postgresql_install 'postgresql' do
         2:    version 13
         3>>   source :nil
         4:    action :install
         5:  end
       
       System Info:
       ------------
       chef_version=17.10.0
       platform=debian
       platform_version=11
       ruby=ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-linux]
       program_name=/opt/cinc/bin/cinc-client
       executable=/opt/cinc/bin/cinc-client
       
       
       Running handlers:
       [2022-12-14T17:36:08+01:00] ERROR: Running exception handlers
       Running handlers complete
       [2022-12-14T17:36:08+01:00] ERROR: Exception handlers complete
       Infra Phase failed. 0 resources updated in 03 seconds
       [2022-12-14T17:36:08+01:00] FATAL: Stacktrace dumped to /tmp/kitchen/cache/cinc-stacktrace.out
       [2022-12-14T17:36:08+01:00] FATAL: ---------------------------------------------------------------------------------------
       [2022-12-14T17:36:08+01:00] FATAL: PLEASE PROVIDE THE CONTENTS OF THE stacktrace.out FILE (above) IF YOU FILE A BUG REPORT
       [2022-12-14T17:36:08+01:00] FATAL: ---------------------------------------------------------------------------------------
       [2022-12-14T17:36:08+01:00] FATAL: Chef::Exceptions::ValidationFailed: Option source must be equal to one of: :repo!  You passed :nil.
```

## :police_car: Expected behavior

Other values than `:repo` should be allowed for the `source`-property.

## :heavy_plus_sign: Additional context

The presence and usage of the `install_method_repo?`-method inside the `postgresql_install`-resource indicates that using the distributions packages is supported.

```ruby
def install_method_repo?
  new_resource.source.eql?(:repo)
end
```

But there is no possibility to trigger this condition as `:repo` is the default as well as the only allowed values.

```ruby
property :source, [String, Symbol],
          default: :repo,
          coerce: proc { |p| p.to_sym },
          equal_to: %i(repo),
          description: 'Installation source'
```

An other value (e.g. `:distro`) should be introduced to reintroduce the `setup_repo`-behavior. A test should be introduced to cover this case.

hamarituc:
# Description

This PR allows to install the native distribution packages of PostgreSQL.

## Issues Resolved

- closes #709

## Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This file is used to list changes made in the last 3 major versions of the postg
 ## Unreleased
 
 - resolved cookstyle error: resources/role.rb:1:1 refactor: `Chef/Deprecations/ResourceWithoutUnifiedTrue`
+- Allow installing distributions native packages
+
 ## 11.0.1 - *2022-12-13*
 
 Standardise files with files in sous-chefs/repo-management
--- a/documentation/postgresql_install.md
+++ b/documentation/postgresql_install.md
@@ -21,7 +21,7 @@
 | ---------------------------------- | ----- | --------------- | ------- | ------------------------------------------------ | -------------- |
 | `sensitive`                        |       | true, false     |         |                                                  |                |
 | `version`                          |       | String, Integer |         | Version to install                               |                |
-| `source`                           |       | String, Symbol  |         | Installation source                              | repo           |
+| `source`                           |       | String, Symbol  |         | Installation source                              | repo, native   |
 | `client_packages`                  |       | String, Array   |         | Client packages to install                       |                |
 | `server_packages`                  |       | String, Array   |         | Server packages to install                       |                |
 | `repo_pgdg`                        |       | true, false     |         | Create pgdg repo                                 |                |
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -193,6 +193,19 @@ suites:
         pg_ver: "13"
     run_list:
       - recipe[test::client_install]
+  - name: client_install_native_13
+    attributes:
+      test:
+        pg_ver: "13"
+    verifier:
+      inspec_tests:
+        - path: test/integration/client_install_native/
+      inputs:
+        pg_ver: "13"
+    run_list:
+      - recipe[test::client_install_native]
+    includes:
+      - debian-11 # Debian 11 only supports PostgreSQL 13 natively
   - name: extension_13
     attributes:
       test:
@@ -242,6 +255,19 @@ suites:
         pg_ver: "13"
     run_list:
       - recipe[test::server_install]
+  - name: server_install_native_13
+    attributes:
+      test:
+        pg_ver: "13"
+    verifier:
+      inspec_tests:
+        - path: test/integration/server_install_native/
+      inputs:
+        pg_ver: "13"
+    run_list:
+      - recipe[test::server_install_native]
+    includes:
+      - debian-11 # Debian 11 only supports PostgreSQL 13 natively
 
   - name: access_12
     attributes:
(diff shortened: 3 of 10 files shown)


bmhughes:
Does this work on RH platforms? As the default packages probably don't work as the OS repo packages don't have a version number in them and I wonder about the configuration paths as well. 

This was reason I shied away from it with the v11 rewrite is that there isn't a super consistent way to figure it out between versions either, I was thinking along the lines of someone who wants to use the native packages wouldn't use the install resource. The config paths probably all want overriding as well.

hamarituc:
> Does this work on RH platforms? As the default packages probably don't work as the OS repo packages don't have a version number in them and I wonder about the configuration paths as well.

You are right. At least under CentOS 7 the installation fails.

> This was reason I shied away from it with the v11 rewrite is that there isn't a super consistent way to figure it out between versions either, I was thinking along the lines of someone who wants to use the native packages wouldn't use the install resource. The config paths probably all want overriding as well.

At least the code to distinguish between the two modes is still present in version 11 (`source`-property, `install_method_repo?`-method). This lead me to the conclusion that the removal of this mode switching function was unintentional. But if it is not intended to support native packages, this code sequences are superflous as well. From my point of view there are two options  to proceed:

1. Support native packages
  * Change `default_client_packages` and `default_server_packages` in such a way they respect the `source`-property when returning the package name.
  * Add a documentation hint that the `version`-property doesn't have an effect if the `source`-property is set to `:native`.

2. Drop support of native packages
  * Remove the `source`-property entirely as there is simply no option to choose.
  * Remove the `install_method_repo?`-method.
  * Add a documentation hint to not to use the `install`-resource for native packages.

What is your preferred option?

bmhughes:
I think option #2 is probably the way to go, I left the stuff in `:install` incase we wanted to come back to it later on but ultimately I think it's going to be awkward to figure out the file paths for the config resources without creating inter-dependencies between the resources which was one of the main goals of v11.

bmhughes:
Hmm having had a look at this, it's going to need quite a bit of work for the cookbook to work with OS packages as the directory detection for config files is broken too. I'll try to take a look at it soon to try and make it nicer as it's override city at the minute to even get close to the run converging.

PR #725, opened 2023-02-20
Use images from GitHub container registry

damacus:
These images should be faster

Signed-off-by: Dan Webb <dan.webb@damacus.io>


damacus:
This surprisingly doesn't seem to make it faster(!)

PR #726, opened 2023-02-21
Fix database ID and CI failure on RHEL

bmhughes:
# Description

Fix database ID and CI failure on RHEL

## Issues Resolved

- n/a

## Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This file is used to list changes made in the last 3 major versions of the postg
 
 ## Unreleased
 
+- Make database names an identifier to allow additional permitted characters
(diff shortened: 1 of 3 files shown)


damacus:
Should we just add the prepare step to the install recipe? Or should we document that the user needs to enable that repo?

bmhughes:
Hmm I think you're right, it's only needed if for the devel package but we already install epel and scl so it's probably a better option.

damacus:
Either way. thanks for doing this!

kitchen-porter:
Released as: [11.2.1](https://github.com/sous-chefs/postgresql/releases/tag/11.2.1)

PR #731, opened 2023-03-24
Fix unparsable pg_hba.conf when a column value is longer than the max width

kristofvandam:
# Description

When a address or hostname is longer than 24 characters, the method is glued to the address in pg_hba.conf.

This would be also a problem for ipv6 addresses.

## Check List


damacus (dismissed):

ramereth (approved):

kitchen-porter:
Released as: [11.2.10](https://github.com/sous-chefs/postgresql/releases/tag/11.2.10)

PR #739, opened 2023-05-03
Make role encrypted_password idempotent

sorcix:
# Description

Setting the `encrypted_password` property on the `postgresql_role` resource always tries to update the password, as the current value is always `nil`. The existing query to the pg_roles table only indicates the presence of a password, not the (encrypted) contents. That's probably the reason it isn't included in the list of values in `load_current_value`.

This change executes a second query only if the encrypted_password property is set, fetching the current value. The `property_is_set?` condition is there because it makes no sense to execute the second query if passwords are not being changed.

## Issues Resolved

## Check List


damacus (approved):

damacus:
Thanks for the clear explanation on the PR here. It made it much easier to review!

damacus:
Right, I've fixed the CI permissions so we're now running cookstyle again. Can you fix those for us, please? (cookstyle -A should fix them all)

sorcix:
Most errors were caused by the unified mode lint. I've put those fixes in a separate commit as it isn't really related to the idempotency change.

ramereth (approved):

kitchen-porter:
Released as: [11.2.11](https://github.com/sous-chefs/postgresql/releases/tag/11.2.11)

PR #742, opened 2023-06-05
OS package support

hamarituc:
# Description

This PR allows to install native PostgreSQL packages from the OS distributors package repository. This PR is based on @bmhughes work. I would like to withdraw #710 in favour of his approach. I tested it against Debian 11 and CentOS 7 and it works as expected. From my point of view it looks ready to merge as it solves #709 and I couldn't identify any issues remaining.

## Issues Resolved

 * #709

## Check List

- [x] A summary of changes made is included in the CHANGELOG under `## Unreleased`
- [x] New functionality includes testing.
- [x] New functionality has been documented in the README if applicable.

PR #743, opened 2023-06-06
OS package install support

bmhughes:
# Description

Add support for OS package installation

Superceeds #742 

## Issues Resolved

- #709 

## Check List

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -28,33 +28,58 @@ module PostgreSQL
 
         raise 'Unable to determine installed PostgreSQL version' if nil_or_empty?(pgsql_package)
 
-        pgsql_package_version = pgsql_package.first[1].fetch('version').to_i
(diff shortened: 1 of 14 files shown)


ramereth (approved):

libraries/helpers.rb
```diff
@@ -73,24 +98,37 @@ def initialized?
       def secure_random
         r = SecureRandom.hex
         Chef::Log.debug "Generated password: #{r}"
+
         r
       end
 
-      def default_server_packages
+      def default_server_packages(version: nil, source: :os)
```
damacus, line 105: We shouldn't be defaulting source type here. That changes the default behaviour right?

bmhughes: It won't change the default because its always specified, but yes we shouldn't be setting a default here you're right.

PR #744, opened 2023-06-16
Allow to configure auth method with complex/multiple option(s)

Annih:
# Description

Some auth methods require complex options, for instance the `ldap` configuration requires either:
- a complex single option `ldapurl`
- several simpler options

Both solutions were not possible with former Regex in the HBA helper lib, as:
- in the single option case, some characters were forbidden
- in the other case, multiple options was just not possible

This PR allows to specify multiple `auth_options` values and to use any non-space characters.

## Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This file is used to list changes made in the last 3 major versions of the postg
 
 ## Unreleased
 
+- Allow to configure auth method with complex/multiple option(s)
+
 ## 11.3.0 - *2023-06-06*
 
 - Allow package installation from OS distribution repository
--- a/libraries/_utils.rb
+++ b/libraries/_utils.rb
@@ -18,7 +18,23 @@
 module PostgreSQL
   module Cookbook
     module Utils
+      AUTH_OPTION_REGEX = /[\w-]+=(?:"[^"]*"|[^\s"]+)/.freeze
+
       private
+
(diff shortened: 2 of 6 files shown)


xorima:
Thank you for this @Annih 
I've popped on a label so the `Release Label Validator` check will pass fyi

Annih:
Thank you @xorima .
Note that I need to fix one thing in the 2 currents commits, then I plan to add 2 new ones (to handle multiple options).
Should be good by the end of the week-end.

Annih:
OK took me several attempts but finally I managed to have the correct behaviour while not breaking existing tests.
I think it is OK for a review, let me know if you need me to change anything!

Annih:
Side note, I noticed the `PgHbaFileEntryAuthOptions` class overrides [.eql?](https://github.com/sous-chefs/postgresql/blob/main/libraries/access.rb#L308), which is OK in most cases but if the other `auth_options` is a `Hash` then it is compared as-is with the current `@options` which is an `Array`.
I didn't want to change that in this PR, due to potential unknown side effects, but IMHO the `Hash` case is useless and the fact no-one complained about the incorrect comparison is a good signal :D

xorima (approved):

kitchen-porter:
Released as: [11.4.0](https://github.com/sous-chefs/postgresql/releases/tag/11.4.0)

Issue #747, opened 2023-06-19
`pg_hba.conf` doesn't reach a stable state


hamarituc:
## :ghost: Brief Description

Using multiple `postgresql_access` resources addressing the same entry, trigger changes in `pg_hba.conf` at each Chef run and thus don't converge to a stable result.

## :pancakes: Cookbook version

11.4.0

## :woman_cook:  Chef-Infra Version

17.10.0

## :tophat: Platform details

Debian 11

## Steps To Reproduce

1. Consider the following recipe

```ruby
postgresql_access 'foo' do
  type 'host'
  database 'foo'
  user 'all'
  auth_method 'md5'
  address '127.0.0.1/32'

  action :delete
end

postgresql_access 'foo' do
  type 'host'
  database 'foo'
  user 'all'
  auth_method 'md5'
  address '127.0.0.1/32'
end
```

2. Run the Chef client. The following line will be inserted into `pg_hba.conf`

```
host    foo                             all                             127.0.0.1/32            md5
```

2. Run the Chef client. The line inserted in step 2 will be removed.
3. Run the Chef client. The line of step 2 is inserted again.
4. Run the Chef client. The line of step 2 is removed again.
5. ...

A similar behavior can be observed in the following recipe.

```ruby
postgresql_access 'bar' do
  type 'host'
  database 'bar'
  user 'all'
  auth_method 'md5'
  address '127.0.0.1/32'
end

postgresql_access 'bar' do
  type 'host'
  database 'bar'
  user 'all'
  auth_method 'scram-sha-256'
  address '127.0.0.1/32'
end
```

In this example the authentication method in the corresponding `pg_hba.conf`-entry will flap between `scram-sha-256` and `md5` with each Chef client run.

## :police_car: Expected behavior

The attributes of the most recent resource definition should win over all earlier definitions.

## :heavy_plus_sign: Additional context

Up to version 10 of the cookbook, the content of `pg_hba.conf` was solely defined by the `postgresql_access`-resource. Since version 11 existing entries will be kept. To deleted unnecessary default entries, the cookbook author has to delete them manually through the `:delete` action.

[...]

PR #748, opened 2023-06-19
Ensure `pg_hba.conf` reaches a stable state

hamarituc:
# Description

The PR solves an issue there `pg_hba.conf` doesn't converge to a stable state.

## Issues Resolved

- #747

## Check List


resources/access.rb
```diff
@@ -80,29 +80,35 @@
...
 
-    if nil_or_empty?(entry)
+  if nil_or_empty?(entry)
+    converge_by("Add grant entry for #{new_resource.type}, #{new_resource.database}, #{new_resource.user}") do
       resource_properties = %i(type database user address auth_method auth_options comment).map { |p| [ p, new_resource.send(p) ] }.to_h.compact
       entry = PostgreSQL::Cookbook::AccessHelpers::PgHba::PgHbaFileEntry.create(**resource_properties)
       config_resource.variables[:pg_hba].add(entry)
-    else
+    end
+  elsif !entry.auth_method.eql?(new_resource.auth_method) || \
```
bmhughes, line 92: You can pass properties to `converge_if_changed` to filter what triggers and update rather than the multiple `if`s. Combining that with `return` rather than `if/elsif` will be much cleaner.

bmhughes:
Hmm from the issue description (#747) I think this needs to be handled in your wrapper cookbook rather than this one as I don't think this can be nicely achieved without breaking the persistent accumulator paradigm and causing resources to be triggered every run.

If this was massaged to make this 'work' it'd result in removing the entry (triggering the first resource convergance) and then adding it back in (triggering another convergance), so it would result in never being able to have a clean run. It should also fail kitchen testing because we can `enforce_idempotency` enabled.

I have a similar base wrapper cookbook and to work with something like this I'd be conditionally declaring the `delete` resource only if a rule didn't exist to (re)create it, I think anything else leaks your implementation into this library cookbook.

Open to thoughts, but from a quick scan through that's where I'm coming from.

hamarituc:
> I have a similar base wrapper cookbook and to work with something like this I'd be conditionally declaring the `delete` resource only if a rule didn't exist to (re)create it, I think anything else leaks your implementation into this library cookbook.

Thanks for the input. I will try this approach.

> Open to thoughts, but from a quick scan through that's where I'm coming from.

After implementing the patch in this PR I realized there is essentially no straightforward way to solve this corner case. I will close this PR therefore.

Just asking out of curiosity: Why the behavior of this resource was changed between version 10 and 11 of the cookbook? I would appreciate the old approach of just defining all HBA rules in the chef recipe. This has the advantage of automatically dropping rules from `pg_hba.conf` not defined anymore. From my point of view this is a bit closer to the chef approach of describing the target state of the system instead of just describing steps which have to be performed.

Is it worth to open a feature request for such a feature? It must not be enabled by default, but just be configurable by a flag for example.

bmhughes:
> Just asking out of curiosity: Why the behavior of this resource was changed between version 10 and 11 of the cookbook? I would appreciate the old approach of just defining all HBA rules in the chef recipe. This has the advantage of automatically dropping rules from `pg_hba.conf` not defined anymore. From my point of view this is a bit closer to the chef approach of describing the target state of the system instead of just describing steps which have to be performed.

It was implemented like this to provide fully featured resources, with the original accumulator-pattern the indiviual resources are 'hollow' and (as far as Chef is concerned) do not do any work. Therefor you can't do things like trigger notifications or subscriptions off of them, nor to do you get the property change notification as the resource isn't aware of what values came before. You only get the update from the underlying template resource.

It has the 'downside' of not automatically removing entries that fall out of scope as the templated file isn't recreated every run, but this can be handled via a cleanup recipe in a wrapper if it is required so I don't think it's a big deal breaker in that respect. 

One option to restore the original behaviour, which should be 'easy' to implement, would be to add a property to the resource to switch off the loading of the existing file from disk, this would restore the original bevahiour without causing too many issues. But it would need to be test. I've done something similar in the auto_accumulator library: https://github.com/sous-chefs/chef_auto_accumulator/blob/30b0cbc26e6906a08f5733b3441d368762641692/libraries/chef_auto_accumulator/resource.rb#L291

PR #765, opened 2023-12-14
Fix library/helpers.rb pulling different package names with multiple PG installations

Justin-Fernbaugh:
# Description

Describe what this change achieves

Currently with multiple PG installations "installed_postgresql_major_version" and "installed_postgresql_package_source" have a chance to pull a different package from each other. The fix is to sort the packages after being filtered such that  calling ".first" always returns the same value.

## Issues Resolved

The functions "installed_postgresql_major_version" and "installed_postgresql_package_source" will maintain the same value.

## Check List


libraries/helpers.rb
```diff
@@ -25,6 +25,7 @@ module Helpers
 
       def installed_postgresql_major_version
         pgsql_package = node['packages'].filter { |p| p.match?(/^postgresql-?(\d+)?$/) }
+        pgsql_package = pgsql_package.sort_by { |key, _values| key[1].to_i }.to_h
```
ramereth, line 28: This doesn't seem to work properly and I'm confused why you're converting that value to an int. If I just change it to the following, it seems to work properly.
```suggestion
        pgsql_package = pgsql_package.sort_by { |key, _values| key }.to_h
```

ramereth (changes requested):

libraries/helpers.rb
bmhughes: I agree with @ramereth here, it doesn't work nor make sense from what I can work out. I'd sort against the `version` key if we wanted to always order by version but that doesn't necessarily help us out.

I'm thinking the better option would be to move the package match logic into a seperate method which `installed_postgresql_major_version` and `installed_postgresql_package_source` could use to get the package so they'll always tie up.

Further to that, I think that matching more than one package is an unknown that we can't handle 100% so it might be better to `raise` if we get more than one package matched.

Justin-Fernbaugh: Really appreciate the suggestions on how to improve this. @bmhughes Moving the matching logic into a function I agree makes a lot of sense here. With that said, the goal of sorting was to select for the newest PG version in order to somewhat accommodate multiple PG installations. 

What are the thoughts on implementing a property overriding the match found from either `installed_postgresql_major_version ` or `installed_postgresql_package_source `. This way those who have more than one PG version on a machine can force one to be used. Also, I'm suggesting to implement this so that upgrades can be performed without having to write custom resource. Thanks for the feedback

ramereth:
@Justin-Fernbaugh can you please update and rebase?

Justin-Fernbaugh:
> @Justin-Fernbaugh can you please update and rebase?

Just pulled and rebased

libraries/helpers.rb
ramereth: @bmhughes As @Justin-Fernbaugh mentioned, this logic is only needed during an in-place upgrade when you will have multiple versions installed. This provides a stable fix so that we can safely do this. This isn't intended to be used long term on a node.

ramereth (dismissed):
@bmhughes let me know if you have any other concerns and what kind of a bump you'd like

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This file is used to list changes made in the last 3 major versions of the postg
 
 ## Unreleased
 
+- Modify installed_postgresql_package_source to get highest PG version from packages
+- Remove PG11 suites from kitchen
+- Remove PG11 testing from CI
```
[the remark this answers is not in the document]
Justin-Fernbaugh: Made these changes and rebased

ramereth (dismissed):

bmhughes:
Sorry I've completely missed this pinging away, let me re-review and try some things out.

libraries/helpers.rb
```diff
@@ -23,12 +23,17 @@ module Cookbook
     module Helpers
       include Utils
 
-      def installed_postgresql_major_version
+      def installed_postgresql_package_logic
```
bmhughes, line 26: Let's refactor this so it's a little clearer:
```ruby
def installed_postgresql_package_logic
  pgsql_package = node['packages'].filter { |p| p.match?(/^postgresql-?(\d+)?$/) }.values

  unless pgsql_package.one?
    Chef::Log.warn("Detected #{pgsql_package.count} installed PostgreSQL versions: #{pgsql_package.keys.join(', ')}. Using latest version.")
    pgsql_package = pgsql_package.max_by { |_key, value| value['version'].to_f }
  end

  raise 'Unable to determine installed PostgreSQL version' if nil_or_empty?(pgsql_package)

  pgsql_package.first
end
```

Justin-Fernbaugh: I was able to implement your refactor/changes with a few modifications

- The Chef Log.warn had to be modified as the '.keys' call no longer existed after extracting the values in line 1. I then modified this to a var which maps over all versions and pulls them together into a single string.
- The '_key' and 'value' separation of the sort had to be reduced to just 'values' as the key no longer exists after calling .values in the first line.
- The 'pgsql_package.max_by' call had to be wrapped in an array to ensure the first value out of the hash is not pulled i.e 'epoch'.

`def installed_postgresql_package_logic
        pgsql_package = node['packages'].filter { |p| p.match?(/^postgresql-?(\d+)?$/) }.values

        unless pgsql_package.one?
          versions = pgsql_package.map { |values| values['version'] }.join(', ')
          Chef::Log.warn("Detected #{pgsql_package.count} installed PostgreSQL versions: #{versions}. Using latest version.")
          pgsql_package = [pgsql_package.max_by { |values| values['version'].to_f }]
        end

        raise 'Unable to determine installed PostgreSQL version' if nil_or_empty?(pgsql_package)

        pgsql_package.first
      end`
      
  Let me know if there's anything else you want modified or more obvious

bmhughes: LGTM!

bmhughes (approved):

bmhughes:
I'm going to merge this but on the whole I'm not 100% sure this is the correct way to go, I thinking it'd be better to `raise` on multiple versions detected and force the user to explicitly specify the version they want to work on rather than have an implicit behaviour that isn't necessarily deterministic.

But let's see how it goes.

kitchen-porter:
Released as: [11.10.0](https://github.com/sous-chefs/postgresql/releases/tag/11.10.0)

PR #782, opened 2025-01-24
Cast `server_config` keys to strings in `postgresql_config` to avoid unnecessary converges

cgunther:
# Description

Reading the existing config on the node in `load_current_value` returns string keys. If you then pass the same configuration but with symbol keys to `server_config` in `postgresql_config`, `converge_if_changed` will believe the resource changed and proceed with converging. The keys are already cast to strings before being passed to the `template` resource, so the template itself doesn't actually change but the `postgresql_config` resource still believes it changed, so any notifications will fire. For example, if you have the resource reload postgresql on change, it'll reload each time the node is converged.

By casting the keys to strings when coercing the property, the resource will no longer believe it changed when the only difference was the class of the key, thus it wont fire notifications unnecessarily.

I'm not very familiar with testing cookbook, so if this change is test-worthy, any points on how to go about this would be much appreciated.

## Issues Resolved

## Check List


damacus:
Hi!

If you could make sure we've got a test that changes that config and it fully works then we're good to go.

cgunther:
Thanks, but I'm not very familiar with how these cookbooks are tested, so any pointers or references I could mimic would be much appreciated so I can add the appropriate coverage.

It looks like the `server_config` property is already covered in two test recipes as a high-level smoke test that it's not completely broken:
https://github.com/sous-chefs/postgresql/blob/0b0da2cd955ab550f8e0408779b926dda908dec1/test/cookbooks/test/recipes/server_install.rb#L9
https://github.com/sous-chefs/postgresql/blob/0b0da2cd955ab550f8e0408779b926dda908dec1/test/cookbooks/test/recipes/server_install_os.rb#L12

In both cases, string keys are passed. Would it be enough to just convert them (or a subset of them) to symbols? Again, that's still largely just a smoke test that it doesn't error given symbol keys.

In a perfect world, I'd imagine the ideal test would be one where the config on disk is set one way before converge, then the resource is called setting it to the same value, but with a symbol key, then asserting the resource didn't change (ie. fire notifications), but I'm not sure exactly what that'd look like, or a reference I can learn from.

I'm also not seeing anywhere that the config is asserted to be sure that each passed value was properly set to approach testing from that perspective. I see a few examples where the `hba` file seems to be tested ([example](https://github.com/sous-chefs/postgresql/blob/0b0da2cd955ab550f8e0408779b926dda908dec1/test/integration/server_install_os/controls/access.rb#L5-L10)), but I can't figure out where that `postgres_hba_conf` variable is coming from to possibly do something similar, as least in how I assume it works, that reads the config on disk after the run and asserts it's values.

bmhughes:
Hi @cgunther!

A test that uses `Symbol` keys should be enough as the kitchen config is set for `multiple_converge: 2` so any changes in the second run will cause a failure.

cgunther:
Got it, thanks for the tip!

I changed the `server_install_os` test recipe to use Symbol keys (`server_install` is still using String keys) and removed my fix to confirm a test would fail due to the change on the second converge. A number of the tests seem to be failing for other reasons, but here's one example that failed as expected:
https://github.com/sous-chefs/postgresql/actions/runs/13380193221/job/37367314945?pr=782

Then I restored my fix and now that same example passes:
https://github.com/sous-chefs/postgresql/actions/runs/13380282223/job/37367541377

bmhughes (approved):

bmhughes:
Looks good @cgunther!

cgunther:
Terrific, thanks for the merge!

PR #786, opened 2025-06-20
allow more mappings per mapname

kristofvandam:
# Description

It is perfectly fine have more mappings per mapname, [ref](https://www.postgresql.org/docs/current/auth-username-maps.html).

The current implementation doesn't allow repetitions of mapnames in `pg_ident.conf`, with this change an entry is unique based on map_name, system_username and database_username.

This allows for multiple system users to be mapped to a single database user.

## Issues Resolved

No issue was created regarding tis PR

## Check List


kristofvandam:
would fix https://github.com/sous-chefs/postgresql/issues/787

damacus (approved):

damacus:
Looks like we got a small test error. Ping me when  this is fixed 👌 

Thanks!

PR #834, opened 2026-08-11
fix: support pg under Habitat Chef

### LINKED ISSUE #833: postgresql_user / postgresql_database fail under Chef 19 (Habitat)
When running Chef 19 packaged as a Habitat package, the `postgresql_user` and `postgresql_database` resources fail with:

```
FATAL: LoadError: pg's C extension failed to load:
  cannot load such file -- 3.4/pg_ext
  libpq.so.5: cannot open shared object file: No such file or directory - /hab/pkgs/core/ruby3_4/3.4.8/.../gems/pg-1.6.3/lib/pg_ext.so
```

### Root Cause

Since v11.0.0, this cookbook installs the `pg` gem to back `postgresql_user` and `postgresql_database`. The native extension `pg_ext.so` links against `libpq.so.5` at runtime.

Habitat packages Chef with library isolation: the Ruby process is started with a scoped `LD_LIBRARY_PATH` pointing only to Habitat-internal paths, and the binary appears to be compiled with `-z nodefaultlib`, which disables both the `ldconfig` cache and default system library paths (`/lib`, `/usr/lib`). System-installed `libpq.so.5` (from `libpq5`) is therefore invisible to the dynamic linker even when correctly installed.

Setting `ENV['LDFLAGS']` or `ENV['DLDFLAGS']` before the gem builds does not help:
- `LDFLAGS` controls executable linking in mkmf, not `.so` linking.
- `DLDFLAGS` is initialised by mkmf from `RbConfig::CONFIG`, not from `ENV`.

Adding system paths to `LD_LIBRARY_PATH` causes a crash (`__nptl_change_stack_perm` symbol error) because the system `libc.so.6` gets loaded alongside Habitat's own bundled libc.

### Environment

- **OS:** Ubuntu 24.04 (noble)
- **Chef version:** 19.3.15 (Habitat)
- **sous-chefs/postgresql:** 13.0.5 (reproduced on all v11+ versions)
- **PostgreSQL:** 16 (PGDG)

### Workaround

Downgrade to Chef 18 (Omnibus). Omnibus uses the system linker and resolves system libraries normally — the issue does not occur.

damacus:
## Summary

- Let RubyGems select the bundled-libpq `pg` platform gem when Chef runs from Habitat, while preserving source builds elsewhere.
- Validate existing Habitat pg installs out of process and replace ruby-platform or non-loadable gems left by an earlier converge.
- Leave healthy Habitat platform gems untouched and cover the shared `postgresql_user` / `postgresql_database` connection path.

Closes #833.

## Verification

- `cookstyle`: 52 files, 0 offenses.
- `env HOME=/private/tmp/postgresql-833-home CHEF_LICENSE=accept-no-persist chef exec rspec`: 73 examples, 0 failures.
- `env KITCHEN_LOCAL_YAML=kitchen.dokken.yml CHEF_LICENSE=accept-no-persist kitchen test access-16-ubuntu-2404`: Ubuntu 24.04 / PostgreSQL 16, 12 controls and 36 assertions passed, including second-converge idempotency.
- Exact Chef Infra Client 19.3.15 Habitat package: an isolated prior `pg-1.6.3` ruby-platform install was rejected, removed by the real `chef_gem` provider, replaced with `pg-1.6.3-aarch64-linux`, and loaded successfully.

## Risks / follow-ups

- The repository's Dokken/Omnibus provisioner cannot install Habitat-based Chef 19 (`chef/chef:19.3.15` is unavailable), so the exact Habitat package upgrade and Ubuntu 24.04 resource integration were validated separately.
- The pg binary gem does not support LDAP-based libpq connection-option lookup. Direct connection parameters and PostgreSQL HBA LDAP configuration are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This file is used to list changes made in the last 3 major versions of the postg
 
 ### Bug Fixes
 
-* Use the pg gem's bundled libpq when Chef runs from Habitat
+* Use the pg gem's bundled libpq when Chef runs from Habitat, replacing incompatible source installs
 * Preserve pre-computed SCRAM-SHA-256 role password verifiers
 
 ## [13.0.5](https://github.com/sous-chefs/postgresql/compare/v13.0.4...v13.0.5) (2026-07-30)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This file is used to list changes made in the last 3 major versions of the postg
 
 ### Bug Fixes
 
+* Use the pg gem's bundled libpq when Chef runs from Habitat
 * Preserve pre-computed SCRAM-SHA-256 role password verifiers
 
 ## [13.0.5](https://github.com/sous-chefs/postgresql/compare/v13.0.4...v13.0.5) (2026-07-30)
--- a/libraries/sql/_connection.rb
+++ b/libraries/sql/_connection.rb
@@ -17,6 +17,7 @@
 
 require_relative '../_utils'
 require_relative '../helpers'
+require 'rbconfig'
 
 module PostgreSQL
   module Cookbook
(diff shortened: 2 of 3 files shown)


damacus:
Closing this in favour of upstream Progress/Chef involvement. The workaround passed local Habitat validation and the standard cookbook matrix, but the affected Chef 19 Habitat runtime cannot be installed or exercised by this repository CI. We will not ship a packaging-specific workaround that we cannot reproduce and maintain in CI. Further tracking remains in #833.

PR #835, opened 2026-08-13
Add Test Kitchen regression test for pg gem rpath under Habitat (#833)

tpowell-progress:
## Summary
Adds a Test Kitchen / InSpec regression control for [#833](https://github.com/sous-chefs/postgresql/issues/833): `postgresql_user`/`postgresql_database` fail under Chef 19+ (Habitat-packaged Chef Infra Client) because the compiled `pg` gem's `pg_ext.so` can't locate `libpq.so.5` at runtime — Habitat's Chef binary runs with a scoped `LD_LIBRARY_PATH` and `-z nodefaultlib`, hiding the system-installed `libpq.so.5` from the dynamic linker.

`kitchen.dokken.yml` already installs Chef 19+ via Habitat by default, and the existing `access` suite already converges `postgresql_user`/`postgresql_database` resources, so no new suite is needed. This PR adds one focused control to `test/integration/access/controls/base_access.rb` that resolves `pg_ext.so`'s path from the running chef-client's own gem environment and runs `ldd` against it, asserting `libpq` resolves. This gives a precise, named failure matching the exact #833 symptom instead of relying on the whole converge aborting less legibly.

## Testing
```
kitchen test access-17
```
Currently red on Habitat-based Chef installs until the rpath fix (baking `-Wl,-rpath` for the postgres lib dir into `pg_gem_build_options`) lands as a follow-up.

This work was completed with AI assistance following Progress AI policies.

### DIFF
--- a/spec/libraries/connection_spec.rb
+++ b/spec/libraries/connection_spec.rb
@@ -1,0 +1,42 @@
+require 'spec_helper'
+require_relative '../../libraries/sql/_connection'
+
+# Regression test for https://github.com/sous-chefs/postgresql/issues/833
+#
+# Under Chef 19 packaged as a Habitat package, the linker is started with a
+# scoped LD_LIBRARY_PATH and `-z nodefaultlib`, so the `pg` gem's compiled
+# `pg_ext.so` cannot find `libpq.so.5` at runtime even when it is installed,
+# because nothing tells the linker where to look besides the default/system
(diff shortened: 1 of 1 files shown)


tpowell-progress:
Was trying to see if this could be surfaced in chef/chef CI and at the last minute Copilot decided sous-chefs/postgresql should be the target

